Issue #77, opened 2014-08-19
Session id parsed from custom HTTP header


alexey-detr:
I've used session middleware to handle sessions in the API consumed by ios-based client. Everything was fine until I started web-based client implementation. I've noticed that cookies are not the best way to work with them in the browsers, there are some restrictions with fetching them in JavaScript. I know that is security reasonable, but anyway I need an ability to pass session id via custom header.

I've forked repository and implemented ability to pass session id via custom header with backward compatible cookies passing. Also added one test to be sure that sessions are loaded via header as well.

Not sure if I should suggest the pull request for this feature. Is this repository about only cookie-based sessions?

PR #79, opened 2014-08-19
Added support for passing session ID using HTTP header

alexey-detr:
These changes add new option `headerName` that can be used to optional passing session ID using custom HTTP header.

Already discussed this issue in #77


Vitaliy-Yarovuy:
+1

alexey-detr:
Guys I have updated the PR. Now it is possible to specify `cookie` option as `null`, so cookies will not be ever sent with responses. Tried to implement it when `cookie` option is not specified at all, but it breaks backward compatibility. Setting `cookie` as `null` allows to use *only* headers to pass session id.

stephenliberty:
So.. when's this getting pulled in?

BonsaiDen:
+1 Would be great to have this in shortly, don't want to re-implement it on my own when it's pretty much already being presented here on a silver platter.

stephenliberty:
FWIW, the express-mysql-session package (and maybe others) specifically look for cookies.. so this doesn't work for them. Issues should be opened on those packages once this finally makes it in..

dougwilson:
FYI for those coming by here waiting for this to be implemented, for now the only work-around is to add the session ID in the `req.signedCookies` object provided by `cookie-parser` module.

hazemsq:
+1 to have this in shortly

simllll:
any news on that?

lime-green:
:+1:

alexprice91:
bump

cojack:
@dougwilson how long we have to wait for you to pull this?

jasonbodily:
Very much interested in this. It's been mentioned this is a 'rare' case. I'm familiar with no other way to authenticate and persist sessions against an API cross domain, as cookies can't be shared. Please merge or address this situation.

dougwilson:
> Please merge or address this situation. 

Just looking through the PR, there was a comment on the name of the option, which was only partially changed; the docs (where the comment still remains) hasn't been updated. Also, the PR currently has conflicts and can't be merged just as a general comment.

Even then, there are multiple threads discussing this topic, at in at least one of them, it was said I would rather accept an PR that allowed a user to read the session ID out in whatever way they choose, instead of still locking them into the decisions of this library. For example, you are locked into Cookies because that's who this module did; this PR would now just lock you into Cookie or a Header, which is not a generally better situation, just a quick Band-Aid that should be better addressed in some PR, please.

kappuccino:
@dougwilson, what are you suggesting ? we should create a middleware just to manipulate this Header ?

alexey-detr:
> this PR would now just lock you into Cookie or a Header, which is not a generally better situation, just a quick Band-Aid that should be better addressed in some PR, please.

Totally agree with @dougwilson. There should be a better more general solution for this problem. For example it could be an optional callback functions specified in configuration that will fetch/save session ID in any way you want. So you will be able to pass session ID in cookies, headers, query string, custom field of JSON-objects in POST request... Not just only cookie or header.

dougwilson (changes requested):
PR needs to be reworked following the discussion in the comments. Also resolve merge conflicts.

simllll:
Just for reference: the above mentioned more general solution can be found in PR #159

README.md
```diff
@@ -28,6 +28,7 @@ Session data is _not_ saved in the cookie itself, just the session ID.
 #### Options
 
   - `name` - cookie name (formerly known as `key`). (default: `'connect.sid'`)
+  - `headerName` - optional HTTP header name to pass session ID, e.g. `X-Session-Token`. (default: `undefined`)
```
billwen: in index.js
var signed = 's:' + signature.sign(val, secret);

secret is an Array, instead of a string, so your code has broken.

azachar:
So what is the progress on this? Thx.

maximelebastard:
Hi, some news about this one ?

witherBattler:
Can someone please check this PR out? It's very needed now since third party cookies are blocked in most browsers...

PR #118, opened 2015-01-29
Update README.md

benheymink:
Added a note for a VERY common use case; using Session in conjunction with PassportJS. In this configuration, Passport always adds an object to the uninitialized session object, causing it to be saved, even if saveUninitialized has been set. It's a very common pitfall, and I spent a while trying to work out why saveUninitialized was not having the effect as described.


dougwilson:
Wow, looking through the Passport repo issues, seems like a lot of people have hit this.

benheymink:
Is it worth me looking at a mod to perhaps add an additional option to play nice with Passport? Or are you happy leaving it with just this PR documenting this side effect?

dougwilson:
The ntoe in the doc is all I'm willing to bend for PassportJS :) Once we start making this library account for other library's bad behavior, it'll never end. I see there are PRs against PassportJS that fix the issue, so I would bet on getting those accepted there.

benheymink:
Good point :-)

PR #119, opened 2015-02-09
Update README.md

jub3i:
spell fix

### DIFF
--- a/README.md
+++ b/README.md
@@ -131,7 +131,7 @@ etc.).
 The default value is `'keep'`.
 
   - `'destroy'` The session will be destroyed (deleted) when the response ends.
-  - `'keep'` The session in the store will be ketp, but modifications made during
+  - `'keep'` The session in the store will be kept, but modifications made during
     the request are ignored and not saved.
 
 #### Cookie options


dougwilson:
Thank you, though I need need to put your name (or something that looks like a name), rather than "eternal student". I can update the commit for you if you like, providing the name :)

jub3i:
thank you for trying to get the attribution correct. you can use any of "jub3i", "jub3i anon", "anon jub3i" or some variation on my profile name like "eternal jub3i" or "jub3i student" - if thats not possible you can just claim this PR as your own :)

i am curious and please excuse my ignorance, but why do you need something that looks like a real name?

dougwilson:
I can use "jub3i", then :)

> i am curious and please excuse my ignorance, but why do you need something that looks like a real name?

The reason comes down to tracking down through commit history, especially for IP issues. I know you didn't touch any code, but I still like to keep the requirement for all commits in general. An example is a user contributes some stolen IP to a project and later a company bring litigation against the project's owners. It's much easier to be able to show no malice from the project if we can track it down.

jub3i:
That is perfect and thank you for the explanation.

dougwilson:
No problem :) See, "jub3i" _looks_ like a pen name, rather than "eternal student", which looks like it's trying to pretend to be a first + last name :)

PR #131, opened 2015-03-04
Change a comment

sungminchang:
from ' // set-cookie ' 
to     ' // set cookie '

Seems to make it clearer what we're doing in the next lines of code after the comment.


dougwilson:
The comment is referring to the name of the header being created in the response.

sungminchang:
thanks for the clarifications and quick reply!

PR #134, opened 2015-03-12
Inserted warning for newbies like me

lindsaymacvean:
I would have found it very useful to read a warning about implementing a production session storage. It took me a few hours of messing round with sessions to realise that I couldn't do it all with this module alone.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,9 +22,10 @@ var session = require('express-session')
 
 Create a session middleware with the given `options`.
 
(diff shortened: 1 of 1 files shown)


dougwilson:
Sounds good. Sorry for the confusion!

lindsaymacvean:
No problem at all :)

gabeio:
I think it should be note also not warning it's not something you need to be warned about it something you do need to note though...

lindsaymacvean:
Makes sense. Changed it.

PR #136, opened 2015-03-14
Cookie-parser no longer dependency

lindsaymacvean:
I believe this statement to be correct. I hope it is ok to suggest edits to the main readme like this :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -171,7 +171,7 @@ By default `cookie.maxAge` is `null`, meaning no "expires" parameter is set
 so the cookie becomes a browser-session cookie. When the user closes the
 browser the cookie (and session) will be removed.
 
-The (cookie-parser)[https://github.com/expressjs/cookie-parser] middleware is no longer a dependency, express-session creates its own cookie, so there is no need to require cookie-parser to support your sessions, and if choose to do so it may cause conflicts.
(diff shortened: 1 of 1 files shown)


dougwilson:
This is a correct statement, though it's (mostly) noted in the version history already. Typically I like to keep the README to be documentation of the version the README is for instead of documentation the history of changes, but I guess I could include this.

So a couple nits:

1. Can you add that this statement applies to versions 1.5.0 and up?
2. I assume you're suggesting this because you ran into this? What is the most common way you are seeing that you need to use `cookie-parser` with this module? (this is just a question for my own curiosity)

lindsaymacvean:
Apologies I dont think this reply posted last night like it should have...

1. Done
2. I did not originally have a need for cookie-parser, it was more that most tutorials in google results for express-session, also used cookie-parser and so they are misleading. Mentioning this in the documentation would just clear up any confusion. (I believe I am not alone with this query). http://stackoverflow.com/questions/27961320/when-should-i-use-cookie-parser-with-express-session 
http://stackoverflow.com/questions/25595873/node-js-express-session-dependency-on-cookieparser

My current use case is just developing a simple restful/hypermedia api for an SPA with an authentication feature using passport.js. Please confirm that I do not need to use cookie-parser to complete this objective? Or better yet point me to documentation that helps me to understand when I would ever need to use cookie-parser in tandem with express-session?

dougwilson:
You never need it :) Your issue was closed because I merged your PR once I saw the update and it's in the README now :) It used to be required back in the really young Express 3.x and lowers days.

Since then, it's considered bad practice for one middleware to mysteriously require you to use another.

PR #137, opened 2015-03-16
Fixed doc about custom store.

nleclerc:
I changed the readme section about custom store implementations because it was incomplete and misleading.
Just added the minimum correct information to get going.

### DIFF
--- a/README.md
+++ b/README.md
@@ -292,7 +292,7 @@ req.session.cookie.maxAge // => 30000
 
 ## Session Store Implementation
 
-Every session store _must_ be an `EventEmitter` and implement the following
+Every session store _must_ extend the `Store` class (`session.Store`) and implement the following
 methods:
 
    - `.get(sid, callback)`


dougwilson:
There is no reason you need to extend the `Store` class; that only makes it easier. It's absolutely not a _must_ like this doc change says.

nleclerc:
I changed it because at first i did create an EventEmitter and if it's
certainly necessary it's clearly not enough.
The doc probably needs to be a bit more substantial then.

nleclerc:
And going that way, there's no obligation to extend EventEmitter either,
just conform to its signature.

dougwilson:
> And going that way, there's no obligation to extend EventEmitter either, just conform to its signature.

We put that there because we reserve the right to start using any EE method on your store, and even static methods like `EventEmitter.listenerCount` on your store.

> The doc probably needs to be a bit more substantial then.

I 100% agree the docs for this need to be _much_ better. Currently there are so many issues coming up here, it's hard to keep up with those and find the time to provide really good docs for creating new stores. A PR with a guide to creating a store (even if the guide itself it in a separate Markdown file), is most welcome.

PR #141, opened 2015-04-08
Don't modify options.secret; copy it to local var

sibartlett:
The options object that is passed in - shouldn't be modified. If the caller is using that options object elsewhere, then we'll break their code. This change broke [Let's Chat](https://sdelements.github.io/lets-chat) today.

I have updated the code so that options.secret is copied to a local variable.


dougwilson:
Published as version 1.11.1. Sorry for the issue this caused, and thanks for making this a PR!

PR #142, opened 2015-04-13
Fix Grammatical error in README.md

zheshishei:
In the _saveUnitialized_ section:

"Choose `false` will also help with race conditions where a client makes multiple parallel requests without a session."

has been changed to 

"Choosing `false` will also help with race conditions where a client makes multiple parallel requests without a session."


dougwilson:
Thank you!

PR #152, opened 2015-05-03
Added mssql-session-store to Compatible Session Stores.

jwathen:


gabeio:
I would like to see the store be a little more mature before we add it to the readme. As it only has 11 commits to it.

dougwilson:
I agree with @gabeio in that this module is very new (it only came into existence a few hours ago, even), so it's hard to judge the quality, but I don't think we can really shut people out of the list unless there is an actual independently-verifiable metric to inclusion, otherwise it's really just a subjective list.

Instead, I would ask: how does this store compare to https://www.npmjs.com/package/connect-mssql ?

Otherwise I'm going to merge this PR unless anyone has a technical objection :)

gabeio:
I see what you mean there is no real way to *test* maturity/quality...

jwathen:
I think it's very similar to connect-mssql which I hadn't found.  I'll collaborate with the author of that package to add some of the functionality that I'd like for the store; I'd rather not split the effort.  I do, however think that it's not a bad idea to get *some* SQL Server store listed so that others don't make the same mistake I did and assume that there's no compatible store.

gabeio:
I wasn't stating that we didn't want mysql in there just no one who knew about that SQL Store asked to add it here it's a community effort :smile:.

dougwilson:
We can simply add both.

PR #154, opened 2015-05-09
Condition "req.session.cookie.expires != null" is not necessary.

exabugs:
Because when "req.session.cookie.expires === null",
cookie don't re-set if you modified cookie.

For example,

1. First
 req.session.cookie.expires === "2015-05-23T14:26:04Z";
 req.session.cookie.path    === "/";

 → Set cookie.

2. Change to
 req.session.cookie.expires === null;
 req.session.cookie.path    === "/admin";

 → Session modified. But don't re-set cookie.

----

Tests is passed.

### DIFF
--- a/index.js
+++ b/index.js
@@ -386,9 +386,7 @@ function session(options){
         return true;
       }
 
-      return cookieId != req.sessionID
-        ? saveUninitializedSession || isModified(req.session)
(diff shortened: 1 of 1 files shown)


dougwilson:
Thank you! The tests only pass because there wasn't a test for this condition. I have added the missing tests and your PR sadly no longer passes since it's removal intentional functionality :(

PR #155, opened 2015-05-10
Add session-rethinkdb to README.md

llambda:
Add session-rethinkdb


gabeio:
:+1: already excited to try this!

llambda:
@gabeio :+1:  Performance appears to be in the same ballpark as Cassandra, Postgres and Mysql. https://github.com/llambda/express-session-benchmarks

PR #164, opened 2015-05-28
Implemented ability to sessionId manually. It fixes #158

### LINKED ISSUE #158: Can't set manual sessionID

stalniy:
Exported cookie sessionId store in case if someone wants to check in few places: cookies, headers, query string, etc. Simplified code and added tests


index.js
```diff
@@ -43,6 +43,74 @@ exports.Store = Store;
...
+    res.setHeader('set-cookie', header.concat(data))
+  },
+
+  /**
+   * Get the session ID cookie from request.
+   *
+   * @return {string}
+   * @private
+   */
+  get: function (req, name, secrets) {
```
dougwilson: Just reference the old `getcookie` function instead of copy and pasting it up here (where it ends up with another level of indentation).

index.js
```diff
@@ -43,6 +43,74 @@ exports.Store = Store;
...
+    return val || undefined;
+  },
+
+  /**
+   * Generate a session ID for a new session.
+   *
+   * @return {String}
+   * @private
+   */
+  generate: function () {
```
dougwilson: Just reference the old `generateSessionId` function instead of copy and pasting it up here (where it ends up with another level of indentation).

index.js
```diff
@@ -43,6 +43,74 @@ exports.Store = Store;
 exports.Cookie = Cookie;
 exports.Session = Session;
 exports.MemoryStore = MemoryStore;
+exports.sessionId = {
+  /**
+   * Set sessionId to cookie.
+   *
+   * @private
+   */
+  set: function (res, name, sessionId, secret, options) {
```
dougwilson: Just reference the old `setcookie` function instead of copy and pasting it up here (where it ends up with another level of indentation).

README.md
```diff
@@ -43,7 +43,31 @@ more information on the different values.
...
+    },
+
+    generate: function(req) {
+      return 'sess:' + req.user.id;
+    }
+  }
+}));
+```
+
+##### genid (deprecated, use sessionId.generate instead)
```
dougwilson: Move the deprecation notice out of the title and into the body.

index.js
```diff
@@ -89,17 +158,25 @@ function session(options){
     , cookie = options.cookie || {}
     , trustProxy = options.proxy
     , storeReady = true
-    , rollingSessions = options.rolling || false;
+    , rollingSessions = options.rolling || false
+    , sessionId = options.sessionId || {};
+
+  sessionId.set = sessionId.set || exports.sessionId.set;
```
dougwilson: We don't want to modify people's incoming objects (`sessionId` in this case); it brings lots of issues.

dougwilson:
This is looking like the most promising PR so far :)

index.js
```diff
@@ -89,17 +158,25 @@ function session(options){
...
+  sessionId.generate = sessionId.generate || options.genid || exports.sessionId.generate;
+
   var resaveSession = options.resave;
   var saveUninitializedSession = options.saveUninitialized;
   var secret = options.secret;
 
-  var generateId = options.genid || generateSessionId;
-
-  if (typeof generateId !== 'function') {
+  if (typeof sessionId.generate !== 'function') {
```
dougwilson: This change will cause a misleading error for the case of `sessionId: {generate: 'foo'}`, because the user didn't pass in the `genid` option and will think the correct solution will be to try `sessionId: {generate: 'foo'}, genid: function(){}`, which wouldn't fix the problem.

stalniy: Didn't get this. But changed the priority of "genid" option. If it's specified `sessionId.generate = options.genid`. Added test for this

stalniy:
Fixed all comments

dougwilson:
Awesome, looks good, though a test is failing with `TypeError: "sessionId.generate" option must be a function` on Travis CI.

stalniy:
Squashed commits and fixed error message. Lets wait for Travis

gabeio:
We should probably leave the note about assuring that the session id's are unique. Cover all bases.

stalniy:
Guys, this lib needs to be completely refactored. It's really hard to understand what it's going on there... 
Move some pissed cookie related peaces under `setcookie`  method. Please review and give me your feedback. I'd like to see this to be released ASAP =)

dougwilson:
FYI @stalniy even if we merge, the release cycle is dependent on the Express 3.x cycle until July comes, so the speed of merging will not help it get released any faster.

dougwilson:
It would also be nice if you commit didn't move around functions for no reason (so the diff is smaller) and we don't want to be too hasty to release something like this, because once it's published, we have to live with it basically forever and can not go back on it. We have made this mistake too many times to want to just release a non-trivial PR in under 24 hours from the pull.

stalniy:
@dougwilson  I see... Will keep in mind your recommendation about functions. Tnx

stalniy:
@dougwilson I moved methods back to their positions. Looks a bit better. Also refactored index.js file and moved some methods under session responsibility. So, currently I see 2 more places which must be simplified (which will help you to support this lib in future):
* `res.end` - I even don't have idea what it's going on there
* `parseOptions` - this function should parse options, set defaults, do checks regarding deprecations. It will decrease amount of closured vars and simplify session middleware

dougwilson:
Cool, but let's not go too crazy on this PR. For example, we cannot add anything to req.session without a major version bump; we have no way to know people are not storing values in those properties right now and we've had bad experience breaking people in the past adding to req.session

stalniy:
Ok, understand. Do you want me to remove last commit?

jwhitmarsh:
What's the latest on this? This is really desirable!

stalniy:
@dougwilson any updates? You that new version will be available in July

stalniy:
@dougwilson is there something I can do In order to speed up merge of this into master?

msyrus:
+1 for this

PR #172, opened 2015-06-18
options add 'nextopen' item

cnwhy:
the default value is `true`.
if `false` and `cookie.maxAge` is any, user close browser then session loss;

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,6 +154,25 @@ The default value is `'keep'`.
   - `'keep'` The session in the store will be kept, but modifications made during
     the request are ignored and not saved.
 
+##### nextopen
+Close the browser to save the session, of course, you must configure the expiration time for the sessionID.  
+
+The default value is `true`.  
+
+  - `false` cookie.maxAge is any, User closes the browser sessionID reset;  
+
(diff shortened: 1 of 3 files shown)


JoeWagner:
Are you trying to accomplish session cookies?  If so you should be able to use the default of `cookie.maxAge === null`.

Also, looks like tests are failing

dougwilson:
So there are a few issues with this PR:

1. Can you explain exactly what this option does?
2. Can you add documentation to the README?
3. Can you add tests?

cnwhy:
Yes, the function has been completed, but did not think of a good name, tentatively named "nextopen";  
When `nextopen == false` and sessionID time has expired, the program will add a no expiration time cookies,it is used to monitor whether the user closes the browser.  
When the user closes the browser, open the browser again to access, will first clear the sessionID, let sessionID re-initialized.

dougwilson:
Thanks for your contribution, but this is better fit as a third-party library than in core. When `maxAge` is `false`, a user is _supposed_ to loose the session when they close the browser, which is extremely useful for people who use computers in libraries or other public places.

PR #175, opened 2015-06-20
Make sessions expirable

halt-hammerzeit:
Adds session timeout parameter

### DIFF
--- a/session/memory.js
+++ b/session/memory.js
@@ -13,6 +13,7 @@
 
 var Store = require('./store')
 var util = require('util')
+var debug = require('debug')('express-session')
 
 /**
  * Shim setImmediate for node.js < 0.10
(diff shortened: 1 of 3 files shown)


dougwilson:
Thank you! Please investigate why the tests don't pass any longer on Travis CI. Also, unless you are OK with this waiting months until a 2.0, please default this to off, as otherwise the timer will be holding the event loop open before when it didn't use to.

kuchumovn:
This thing seems glad, alas!
I fed it enough code

session/memory.js
```diff
@@ -32,12 +33,47 @@ module.exports = MemoryStore
...
+    var self = this
+    this.session_timeout_checker = setInterval(function() {
+      debug('checking sessions for timeout')
+      for (var sessionId in self.session_timers) {
+        if (self.session_timers[sessionId] === 0) {
+          debug('session ' + sessionId + ' timed out, destroying it')
+          self.destroy(sessionId)
+        }
+        else {
+          self.session_timers[sessionId]--
```
saperal: The timer logic is inaccurate by a minute.
If the timer is 0 at this time. The session needs to be cleaned up. Instead it will wait another minute.

halt-hammerzeit: @saperal indeed, `setInterval` doesn't fire immediately, therefore it should be

```javascript
+      for (var sessionId in self.session_timers) {
+        self.session_timers[sessionId]--
+        if (self.session_timers[sessionId] <= 0) {
+          debug('session ' + sessionId + ' timed out, destroying it')
+          self.destroy(sessionId)
+        }
```

Seems that noone is interested in merging this PR tho (it has already been 3 months).

dougwilson:
Please remove the use of `sinon` and make this more efficient, such that a timer is not fired every single minute, even when there is nothing to expire.

halt-hammerzeit:
Oh really, so this place isn't deserted then?

I don't know how can `sinon` be removed because it provides testing of `setTimeout`s

dougwilson:
> I don't know how can sinon be removed because it provides testing of setTimeouts

Not sure, but Sinon.JS is about to drop Node.js 0.8 support (from reading https://github.com/sinonjs/sinon/pull/859) and this module has no timeline to stop supporting Node.js 0.8, so we need our tests to run and dependencies to install on Node.js 0.8.

halt-hammerzeit:
@dougwilson currently I no longer have interest in pushing forward this PR. It was relevant for me at the time I submitted it, and now I've moved on to Koa. I can use this code as you wish.

dougwilson:
Ok, I''ll close the PR for now :) If someone comes by & is interested, they can always pick it back up.

PR #179, opened 2015-06-23
add cookieFn to dynamically create cookie options depending on request

jpospychala:
I have a backend that runs on more than one domain and depending from which domain request comes from, I'd like the session cookie to point to that domain.

Currently this is not possible, because you can only specify one domain in cookie.domain option.

One way to solve this problem is to have a configurable function that is called for each request to produce cookie options. This is what my pull-request adds. I'd be happy to hear your comments.

### DIFF
--- a/index.js
+++ b/index.js
@@ -68,6 +68,7 @@ var defer = typeof setImmediate === 'function'
  *
  * @param {Object} [options]
  * @param {Object} [options.cookie] Options for cookie
+ * @param {Function} [options.cookieFn] Function producing cookie options per request
  * @param {Function} [options.genid]
  * @param {String} [options.name=connect.sid] Session ID cookie name
  * @param {Boolean} [options.proxy]
@@ -87,6 +88,7 @@ function session(options){
     , name = options.name || options.key || 'connect.sid'
     , store = options.store || new MemoryStore
     , cookie = options.cookie || {}
+    , cookieFn = options.cookieFn
     , trustProxy = options.proxy
     , storeReady = true
     , rollingSessions = options.rolling || false;
@@ -98,6 +100,10 @@ function session(options){
 
   if (typeof generateId !== 'function') {
     throw new TypeError('genid option must be a function');
+  }
+
+  if (cookieFn && typeof cookieFn !== 'function') {
+    throw new TypeError('cookieFn option must be a function');
   }
 
   if (resaveSession === undefined) {
@@ -139,7 +145,7 @@ function session(options){
   store.generate = function(req){
     req.sessionID = generateId(req);
     req.session = new Session(req);
-    req.session.cookie = new Cookie(cookie);
+    req.session.cookie = new Cookie(cookieFn && cookieFn(req) || cookie);
   };
 
   var storeImplementsTouch = typeof store.touch === 'function';
--- a/test/session.js
+++ b/test/session.js
@@ -638,6 +638,21 @@ describe('session()', function(){
     })
   })
 
+  describe('cookieFn option', function(){
+    it('should reject non-function values', function(){
+      assert.throws(session.bind(null, { cookieFn: 'bogus!' }), /cookieFn.*must/)
+    });
+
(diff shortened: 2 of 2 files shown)


dougwilson:
Why not just leave out the domain options? Then the cookie just defaults to the current domain.

jpospychala:
Actually my setup is two totally different domains (e.g. a.com and b.com) with their subdomains (main.a.com, other.a.com, main.b.com, other.b.com). We wanted cookies within domain to be seen by all subdomains (main.a and other.a.com sets a.com cookie, main.b.com and other.b.com sets b.com cookie)

jpospychala:
I saw something similar asked on stackoverflow: http://stackoverflow.com/questions/19104292/express-session-with-different-cookie-domain-per-request with the answer being: you should have as many session middleware instances as domains+session stores, but this makes it really hard to further maintain user sessions.

dougwilson:
Gotcha. The biggest issue I see with this PR is this approach will mean you'll have to make a PR to every module that does cookies...

dougwilson:
A more generic solution would be to use the `vhost` middleware wrapping just this middleware, to me.

dougwilson:
```js
var sessionA = expressSession({ /* arguments for site A */ })
var sessionB = expressSession({ /* arguments for site B */ })

// ... now in your middlewares

app.use(vhost('*.a.com', sessionA))
app.use(vhost('*.b.com', sessionB))

// ... continue
```

dougwilson:
> you should have as many session middleware instances as domains+session stores, but this makes it really hard to further maintain user sessions.

How does it make it really hard to further maintain user sessions? You should be pointing them all to the same `store` instance, so it's no different than if you where using Node.js `cluster` module or running your server on multiple hosts with a load balancer.

jpospychala:
I saw that session middleware sets generate function (that uses cookie options) on store object so my understanding was that you have to use separate store for each session, otherwise two session middlewares would overwrite their store.generate function (and only cookie options from last session middleware would be used).

jpospychala:
RE why harder - in my project, we also use websockets, and on specific message from frontend, backend resets user session by calling sessionStore.client.EXPIRE(session_id, ttl). Now having two session middlewares, each with separate store, this would become really hacky. I would either have to dispatch to correct store or assume that both stores are same. (This is still under my assumption that each session middleware needs separate store instance in order to not overwrite their store.generate functions)

jpospychala:
I don't quite understand why I would have to make PRs to all cookie modules. My new option seemed to me optional and not braking any existing behaviors.

Anyway I'm now leaning towards having two middleware instances (each for each vhost as in your example above), each having separate store but with same settings, so I'll be using single store to remove all sessions anyway.

So I guess that this PR can be discarded.

dougwilson:
> I don't quite understand why I would have to make PRs to all cookie modules. My new option seemed to me optional and not braking any existing behaviors.

This was just to highlight the issue: that if you have this issue with this module, you're going to run into this issue with any other module you use that sets cookies, as they all take just a static set of options.

All I'm doing is trying to help you figure out a _generic solution that works for any module similar to this module_. If it can be solved without modifications here, then that's what we prefer, as otherwise, there is always going to be something to add here that makes for a quick and easy solution, but leads to module code bloat, more bugs, and harder maintenance is all.

dougwilson:
> I saw that session middleware sets generate function (that uses cookie options) on store object so my understanding was that you have to use separate store for each session, otherwise two session middlewares would overwrite their store.generate function (and only cookie options from last session middleware would be used).

Your statement here, though, highlights an issue in this module: yes, apparently it touches he store, to which I can only ponder how that even got in here and just bed sad :(

You could probably work around that by using `Object.create`:
```js
var store = ...
var sesionA = expressSession({ store: Object.create(store), /* options */ })
var sesionB = expressSession({ store: Object.create(store), /* options */ })
```

That way the `.generate` does not actually get set onto the underlying `store` object.

jpospychala:
ah good one... thanks! I'll close this PR...

PR #182, opened 2015-07-02
Add check IP option

vitaliy-orlov:
It adds the ability to match IP with IP client, which is stored initially in the session.It allows you to protect against theft of cookies and use its attacker.

### DIFF
--- a/index.js
+++ b/index.js
@@ -89,7 +89,8 @@ function session(options){
     , cookie = options.cookie || {}
     , trustProxy = options.proxy
     , storeReady = true
-    , rollingSessions = options.rolling || false;
+    , rollingSessions = options.rolling || false,
+    , checkIP = options.checkIP || false;
   var resaveSession = options.resave;
   var saveUninitializedSession = options.saveUninitialized;
   var secret = options.secret;
@@ -140,6 +141,10 @@ function session(options){
     req.sessionID = generateId(req);
     req.session = new Session(req);
     req.session.cookie = new Cookie(cookie);
+    
+    if (checkIP) {
+      req.session.clientIP = req.headers['x-forwarded-for'] || req.connection.remoteAddress;
+    }
   };
 
   var storeImplementsTouch = typeof store.touch === 'function';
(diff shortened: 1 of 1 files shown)


dougwilson:
A big :-1: from me, as this can be done entirely from user code without library modification. The implementation is also very insecure, as i can easily spoof a request header.

vitaliy-orlov:
As far as I know, the header will be overwritten by the proxy server, if you set the header manually.

gabeio:
That is if you are running it behind a reverse proxy & that the proxy is parsing out/for the forwarded for/ correcting the forwarded for.

gabeio:
Also I just want to point out directly saving IP addresses is dangerous. If someone were to break into your database they not only would have all your client info but now they also have IP addresses associated with your clients. It would be *far* safer to securely hash them if you were to store them at all.

vitaliy-orlov:
Yeah, so dangerous...

But if someone were to break into my database, IP addresses could not be so sweet unlike table users.

I agree with you, that it is dangerous in some situation, but it also can help.

JoeWagner:
>this can be done entirely from user code without library modification. The implementation is also very insecure, as i can easily spoof a request header.

Second that

Also, If I move coffee shops or my ISP reconfigures my IP, my session is gone.  In my opinion not a very useful feature.
:-1:

gabeio:
There is no situation where you need the users ip vs a hash of it which can
authenticate their ip.

dougwilson:
Closing in favor of outside-this-module-implementation (you can even publish this as an npm module!). The following is your implementation made into a npm module, but you can implement a simpler version if you don't want to make it look like an npm module.
```js
var createError = require('http-errors')
var expressSession = require('express-session')

module.exports = function (options) {
  var opts = options || {}

  // get the store
  var store = Object.create(opts.store || new expressSession.MemoryStore())
  var gen = store.generate

  // add to session generation
  store.generate = function (req) {
    gen(req)
    req.session.clientIP = req.ip
  }

  opts.store = store

  var middleware = expressSession(opts)

  return function (req, res, next) {
    middleware(req, res, function (err) {
      if (err) return next(err)
      if (req.ip === req.session.clientIP) next()
      next(createError(403, 'Incorrect client IP address of the cookie', {
        code: 'EINCIP'
      }))
    })
  }
}
```

This is just a straight implementation taken from your PR, but with `req.headers['x-forwarded-for'] || req.connection.remoteAddress` replaces with Express's `req.ip`, which takes proxies into consideration. After writing that, there leaves a lot to be desired, which would mean adding this would just result in a whole bunch of feature requests/bug reports for it in the long run.

Publish to npm as like `express-session-secureip` or something :)!

vitaliy-orlov:
Thanks for your reply :)

PR #187, opened 2015-07-26
add cluster-store to compatible stores list

coolaj86:
A wrapper for using in-process / embedded stores - such as SQLite (via knex), leveldb, files, or memory - with node cluster (desirable for raspberry pi 2 and other multi-core embedded devices).


dougwilson:
Thank you! Sorry I forgot to actually push the merge from a while ago :)

PR #195, opened 2015-08-11
Add connect-session-sequelize to stores list

hogart:


gabeio:
Just to let you know http://sequelize.org/ seems to be offline...

PR #209, opened 2015-09-15
Autoset secure cookie

fracmak:
Added ability to automatically set the secure cookie value based on the trust proxy settings when autoSecure option is set.

This is useful when you're working in mixed environments where you're developing locally in an insecure environment, but the higher environments are secure and you want to trust express's request.secure option for scoping your cookies.

### DIFF
--- a/index.js
+++ b/index.js
@@ -88,6 +88,7 @@ function session(options){
     , store = options.store || new MemoryStore
     , cookie = options.cookie || {}
     , trustProxy = options.proxy
+    , autoSecure = options.autoSecure || false
     , storeReady = true
     , rollingSessions = options.rolling || false;
   var resaveSession = options.resave;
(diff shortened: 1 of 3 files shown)


dougwilson:
Hi! Can you add some documentation to the readme as well as make this opt-in behavior?

index.js
```diff
@@ -139,6 +139,7 @@ function session(options){
   store.generate = function(req){
     req.sessionID = generateId(req);
     req.session = new Session(req);
+    cookie.secure = cookie.secure || issecure(req, trustProxy);
```
dougwilson: This will never work when serving requests on parallel both secure and not, because you are altering global state that belongs to multiple requests at the same time. This will have to be fixed before we can accept the PR.

dougwilson:
Once the two issues are addressed, I think this is useful functionality that we can merge!

index.js
fracmak: ah, good catch, let me fix that

fracmak:
issues addressed

dougwilson:
Thanks again! This has just been merged with a slight modification of simplifying the API to just be `{ cookie: { secure: 'auto' } }`.

PR #219, opened 2015-10-28
Update README.md

louischatriot:
Added `express-nedb-session` to the list of session stores. It uses (as it name states) NeDB for data persistence.

### DIFF
--- a/README.md
+++ b/README.md
@@ -416,6 +416,9 @@ and other multi-core embedded devices).
 [connect-mongo-url]: https://www.npmjs.com/package/connect-mongo
 [connect-mongo-image]: https://img.shields.io/github/stars/kcbanner/connect-mongo.svg?label=%E2%98%85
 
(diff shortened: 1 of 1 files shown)


dougwilson:
I know it's your module and all, but I'm slightly hesitant to add it, as the very tops says:

> this library is not maintained anymore, please don't submit anymore pull requests except for bugfixes

Should people still be using it? Is there some kind of replacement people should go for instead?

louischatriot:
Right, the phrasing is not very clear. I mean that I'll not add any feature but still take bug reports. That's a warning I copy pasted on lots of projects without rephrasing it :) I could say it's feature-freezed instead if you prefer (not like there is any feature to add anyway). In any case the underlying db is an active project.

dougwilson:
Thanks, makes sense! You don't need to change the wording on your module, I just wanted to verify :) I have merged the PR.

louischatriot:
Got it, thanks !

PR #224, opened 2015-11-05
Add optional callback to reload method

FelipeBarrosCruz:
Fix the pattern in metod destroy and regenerate.
Add optional callback to reload method.


dougwilson:
In general, we are trying to move in the opposite direction: we don't want optional callbacks to exists and the ones that do are for backwards-compatibility reasons.

PR #238, opened 2015-12-04
make sure cookie is not undefined

marcushat:
Was having sess.cookie show up undefined sometimes, and it was crashing my server when it tried to read the expire property. This doesn't fix the root of the bug but it does stop the process from crashing in the meantime. This error was also fixed by clearing my browser cache. The client side shouldn't be able to crash the process.

### DIFF
--- a/session/store.js
+++ b/session/store.js
@@ -74,6 +74,10 @@ Store.prototype.load = function(sid, fn){
  */
 
 Store.prototype.createSession = function(req, sess){
+  if(typeof sess.cookie == 'undefined') {
+  	sess.cookie = {};
+  }
+
   var expires = sess.cookie.expires
     , orig = sess.cookie.originalMaxAge;
   sess.cookie = new Cookie(sess.cookie);


dougwilson:
Thanks! Please add a test that fails without your change and succeeds after your change.

marcushat:
New to node, don't know what's causing it to be undefined just know that it definitely shouldn't be. Hopefully someone else can find the root cause. I'm not doing anything with cookies in my app and I'm using the default session store when this happens.

dougwilson:
Is there a way you can provide me full code, versions, and instructions on how to reproduce the issue? I can then debug it and try to figure it out, if you want this to be merged. Without a test, we can't actually merge anything, as 1) we're not sure if this is the actual solution and 2) we can accidentally revert the change without realizing it, due to it not being tested (this is why the PR is failing the automatic checks).

marcushat:
I'll send you a link to a private repo of the code in a bit.

dougwilson:
Hi @marcushat , any update to getting tests added to this PR so I can merge? We're going to cut a release pretty soon so wanted to check in.

dougwilson:
Hi @marcushat , I'm going to close this because there hasn't been a reply in almost a month now. If you want to get tests added in, please feel free to make another pull request with everything!

PR #240, opened 2015-12-10
Prevent the `rolling: true` option from overriding the `saveUnitialized: false` option

JamesMGreene:
Prevent the `rolling: true` option from overriding the `saveUnitialized: false` option.

Fixes #239.

PR #242, opened 2015-12-10
Expire cookie immediately upon destroying session

### LINKED ISSUE #241: Cookie should expire immediately when session is destroyed
I believe that the session cookie should be forced to immediately expire when the session is destroyed on the server-side.

This keeps things much cleaner, especially when combined with my changes in PR #240 to fix the interaction between `rolling: true` and `saveUnitialized: false`.

JamesMGreene:
Force an existing session's cookie to expire immediately upon destroying the session.

Fixes #241.

### DIFF
--- a/index.js
+++ b/index.js
@@ -157,11 +157,7 @@ function session(options) {
   store.generate = function(req){
     req.sessionID = generateId(req);
     req.session = new Session(req);
-    req.session.cookie = new Cookie(cookieOptions);
-
-    if (cookieOptions.secure === 'auto') {
-      req.session.cookie.secure = issecure(req, trustProxy);
-    }
+    req.session.cookie = createCookie(cookieOptions, req, trustProxy);
   };
 
   var storeImplementsTouch = typeof store.touch === 'function';
@@ -217,29 +213,47 @@ function session(options) {
 
     // set-cookie
     onHeaders(res, function(){
+
+      // Is this an existing session that is being destroyed?
+      var isDestroying = isBeingDestroyed(req);
+
       if (!req.session) {
         debug('no session');
-        return;
+        if (!isDestroying || isInitialSession(req)) {
+          return;
+        }
       }
 
-      if (!shouldSetCookie(req)) {
-        return;
+      var cookie = req.session ? req.session.cookie : undefined;
+      if (isDestroying) {
+        if (cookie == null) {
+          debug('creating expired cookie');
+          cookie = createCookie(cookieOptions, req, trustProxy);
+        }
+
+        // Set the cookie to immediately expire on the client
+        cookie.maxAge = 0;
       }
 
       // only send secure cookies via https
-      if (req.session.cookie.secure && !issecure(req, trustProxy)) {
+      if (cookie.secure && !issecure(req, trustProxy)) {
         debug('not secured');
         return;
       }
 
-      if (!touched) {
-        // touch session
-        req.session.touch()
-        touched = true
+      if (!isDestroying) {
+        if (!shouldSetCookie(req)) {
+          return;
+        }
+        else if (!touched) {
+          // touch session
+          req.session.touch();
+          touched = true;
+        }
       }
 
       // set cookie
-      setcookie(res, name, req.sessionID, secrets[0], req.session.cookie.data);
+      setcookie(res, name, req.sessionID, secrets[0], cookie.data);
     });
 
     // proxy end() to commit the session
@@ -396,6 +410,16 @@ function session(options) {
       });
     }
 
+    // check if session has been or is being destroyed
+    function isBeingDestroyed(req) {
+      return req.sessionID && req.session == null;
+    }
+
+    // check if session is the initial session
+    function isInitialSession(req) {
+      return req.sessionID && cookieId != req.sessionID;
+    }
+
     // check if session has been modified
     function isModified(sess) {
       return originalId !== sess.id || originalHash !== hash(sess);
@@ -437,8 +461,8 @@ function session(options) {
 
     // determine if cookie should be set on response
     function shouldSetCookie(req) {
-      // cannot set cookie without a session ID
-      if (typeof req.sessionID !== 'string') {
+      // cannot set cookie without a session ID and a session
+      if (typeof req.sessionID !== 'string' || !req.session) {
         return false;
       }
 
@@ -490,6 +514,25 @@ function session(options) {
     });
   };
 };
+
+/**
+ * Create a new Cookie for this request.
+ *
+ * @param {Object} cookieOptions
+ * @param {Object} req
+ * @param {Boolean} [trustProxy]
+ * @return {Cookie}
+ * @private
+ */
+function createCookie(cookieOptions, req, trustProxy) {
+  var cookieOpts = cookieOptions || {};
+
+  var cookie = new Cookie(cookieOpts);
+  if (cookieOpts.secure === 'auto') {
+    cookie.secure = issecure(req, trustProxy);
+  }
+  return cookie;
+}
 
 /**
  * Generate a session ID for a new session.
--- a/test/session.js
+++ b/test/session.js
@@ -1,4 +1,3 @@
-
 process.env.NO_DEPRECATION = 'express-session';
 
 var after = require('after')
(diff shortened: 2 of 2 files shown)


JamesMGreene:
@dougwilson: I'm sure you're day-job-and-open-source-swamped like the rest of us but if there is any chance of getting PR #240 and this PR merged and released with the next 2 weeks or so, my team and I would appreciate it! :+1: :gift_heart:

dougwilson:
Hi James, sorry I haven't been on a computer since you made this. I will look at this pull request as soon as I can. I cannot comment on when this pull request would be accepted, because I need to understand first what the pull requests does and understand if this is a desired feature, if there are edges cases, etc.

I know you feel like it is a bug, but I need to look at it more, as to me, if feels like either a bug or a new feature.

JamesMGreene:
Sounds good, @dougwilson. Thanks! :+1:

If you do end up feeling it is a new feature and would rather implement it as an option rather than as the new default behavior, I'd be fine with refactoring it to work as such if you give me a bit of design suggestion/guidance on how you'd like it to work.

I'll keep an eye out for any feedback requiring changes on my part. :eyes:

JamesMGreene:
@dougwilson: As mentioned in the issue (#241):
> Perhaps this should only be the expected behavior if the cookie would've been updated anyway, e.g. if the `rolling: true` option was enabled.

JamesMGreene:
@dougwilson: If there is any chance of this being merged and released before Monday, I'd :yellow_heart: : you for it.

dougwilson:
Hi @JamesMGreene , I'm sorry, I really wanted to (even did a few commits to this repo tonight, yay!) get this done tonight, before Monday for you. I just didn't get to, but what I did do is at least published the changes to npm that we had so far, and then I really want to get this change done on Monday for you.

JamesMGreene:
@dougwilson: If it makes it in this week, I can probably still pull it into our release branch. Monday would be best if possible so I can just get it into the mainline branch, though. :+1:

dougwilson:
Yea, I'm working on it. Just looked at it and there are all kinds of nits, like the function in `index.js` is not alphabetically sorted like the rest, the tests are in the wrong location/their descriptions are too long. I'm working on adjusting.

dougwilson:
And a reassignment of an argument, which results in a v8 de-opt.

dougwilson:
Hi! It seems like if a user calls `req.session.destroy()` in their code, then it won't actually set the expired cookie. This seems very unexpected and one would expect that the method of destruction would not change if the cookie is expired or not. Example test app:
```js
var app = express()
.use(session({ secret: 'keyboard cat', cookie: { maxAge: 86400000 } }))
.use(function(req, res, next){
  req.session.count = req.session.count || 0;
  req.session.count++;
  if (req.session.count === 2) {
    req.session.destroy(function(err) {
      if (err) return next(err);
      res.end();
    });
  } else {
    res.end();
  }
});
```

index.js
```diff
@@ -571,6 +582,25 @@ function issecure(req, trustProxy) {
...
 /**
+ * Create a new Cookie for this request.
+ *
+ * @param {Object} cookieOptions
+ * @param {Object} req
+ * @param {Boolean} [trustProxy]
+ * @return {Cookie}
+ * @private
+ */
+function createCookie(cookieOptions, req, trustProxy) {
```
dougwilson: Can you move this to be alphabetical within all these functions while you're working on the destroy issue?

index.js
```diff
@@ -571,6 +582,25 @@ function issecure(req, trustProxy) {
...
+  cookieOptions = cookieOptions || {};
```
dougwilson: Please don't re-assign an argument variable, as v8 will never optimize the function.

dougwilson:
Hi @JamesMGreene , I'm not in a rush, but I know you seemed to be, so I just wanted to put a note here just in case you missed my comment in https://github.com/expressjs/session/pull/242#issuecomment-170593341

JamesMGreene:
Thanks, @dougwilson. You caught me in the thralls of our busy release sprint, so I didn't have any time to address this.

I should be able to take a look this week.  I'm OK with this change not going out with our next release; my PR #240 being merged into `v1.13.0` was the more important change.  :+1:

index.js
scurker: @dougwilson I don't think this is true, it appears to only be applicable when [mentioning arguments](https://github.com/petkaantonov/bluebird/wiki/Optimization-killers#3-managing-arguments). If you have another source, I'd be curious to see it.

JoeWagner: @scurker What @dougwilson says above is correct.  [Here](https://gist.github.com/JoeWagner/d89ff977e47e039387f0) is a gist demonstrating the optimization pitfall

scurker: @JoeWagner Thanks for the heads up! I learned something new today.

index.js
JamesMGreene: Done.

JamesMGreene:
> the tests are in the wrong location/their descriptions are too long

@dougwilson: Can you fill me in more on this comment?

index.js
JamesMGreene: Fixed.

JamesMGreene:
> It seems like if a user calls req.session.destroy() in their code, then it won't actually set the expired cookie. This seems very unexpected and one would expect that the method of destruction would not change if the cookie is expired or not.

Working through this issue now.

JamesMGreene:
Alright, so the problem with when the user calls `req.session.destroy()` is that `shouldDestroy(req)` returns `false`, ergo the `isBeingDestroyed` variable is also set to `false`:

```js
// Is this an existing session that is being destroyed?
var isBeingDestroyed = !!cookieId && shouldDestroy(req);
```

@dougwilson et al: Any advice on how to workaround this situation?

dougwilson:
Hi @JamesMGreene, sorry I didn't reply right away; I was just trying to focus on pushing out Express 4.14 :) So I think that that is probably just not a full understanding of the current code, that that's OK! At least how it is designed right now, the `shouldDestroy` and similar is only there to track if the "auto saving" part of this module should basically destroy it instead of save it. This functionality is completely independent of tracking if the session has been manually destroyed or not. You probably have to add manual tracking for this, similar to how we manually track if a session has manually been saved or not.

JamesMGreene:
OK, gotcha @dougwilson. I'll look into that further.

JamesMGreene:
@dougwilson: Finally found a bit of time to dig into this further and get it all working.  Please re-review the new version when you get a chance.  Thanks!

dougwilson:
Hi @JamesMGreene , it looks like the last commit was Dec 10, 2015. Is that the most recent version?

JamesMGreene:
@dougwilson: No, sorry, that was because I did a `git commit --amend` (so it maintained the old commit's date).  I've pushed a new replacement commit now dated as today for better accuracy.

JamesMGreene:
_\*gently pokes @dougwilson\*_ No rush but a gentle reminder. :)

bsodmike:
@JamesMGreene You'll need to clean up re. the conflict 👍

JamesMGreene:
Rebased and dealt with merge conflicts again. :+1:

dougwilson (approved):

index.js
```diff
@@ -396,6 +410,16 @@ function session(options) {
       });
     }
 
+    // check if session has been or is being destroyed
+    function isBeingDestroyed(req) {
+      return req.sessionID && req.session == null;
```
dougwilson: Should this check that `unsetDestroy` is true here? That option is what makes the unsetting of `req.session` actually mean the session is to be destroyed, otherwise it does nothing.

test/session.js
```diff
@@ -1269,14 +1532,14 @@ describe('session()', function(){
       .get('/')
       .expect(shouldNotHaveHeader('Set-Cookie'))
       .expect(200, function(err, res){
-        if (err) return done(err);
```
dougwilson: Looks like a bunch of accidental changes here with the formatting; I can fix, but figured I'd comment here in case you had to make a change from the question above.

test/session.js
```diff
@@ -1258,8 +1257,272 @@ describe('session()', function(){
...
+          expiry = expires(res)
+          assert.ok(expiry)
+
+          var b = new Date(expiry)
+
+          delta = b.valueOf() - a.valueOf()
+          assert.ok(delta <= 0)
+
+          delta = b.valueOf() - now.valueOf()
+          assert.ok(delta <= 0)
```
dougwilson: There is a `shouldSetCookieToExpireIn` helper that could probably be used here if you want.

index.js
dougwilson: Hmm, if I add that check here, the tests don't pass any longer, but if I don't add the check, then `req.session = null` will destroy the cookie even if the user doesn't want `req.session = null` to destroy the session...

JamesMGreene: Hmm, sounds like I must need another unit test for that last condition you mentioned. I'll take a peek when I get a chance.

test/session.js
JamesMGreene: Yeah, that doesn't surprise me too much as it seemed like someone changed the test code style (e.g. to be semi-colon-less) in the time since my initial PR submission. I'll try to clean them up when I make my updates.

test/session.js
JamesMGreene: Thanks for letting me know.  I'll take a peek and see if/where it makes sense to use.

dougwilson:
Hi @JamesMGreene , I know it has been a long time. I let these PRs just sit and I'm trying to make it right by getting them all merged in. I went and rebased all your changes, and even made the adjustments I noted above. It doesn't seem like I'm able to push to your fork, though.

If you haven't already done so, checking off the allow on the right hand side (https://help.github.com/en/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork) would allow me to push up the changes + rebase and then merge the pull request.

But this is not a requirement to land your changes, it just will make the PR show as "merged" instead of "closed". If you're no longer around to do so / don't want to / whatever, it's no big deal. I will plan to merge it within a couple weeks if I don't hear back.

bjohansebas:
closes in favor of #1138

PR #244, opened 2015-12-11
BUG: Add test for incorrect Content-Length with async touch

gobengo:
Add a test case for a bug I found in the wild when using express-session with Error middlewares. It results in the http socket being destroyed before the response body is fully written.

Because of this response body slicing
https://github.com/expressjs/session/commit/5c503e7552ea48342f4ce4ce1322606eb8ee217c#diff-168726dbe96b3ce427e7fedce31bb0bcR226

And then finalhandler destroys the `req.socket`, which `=== res.socket`
https://github.com/pillarjs/finalhandler/blob/master/index.js#L104

And so then the response can't get that last byte to the client, and the Content-Length is then off by one

And Chrome complains: `Failed to load resource: net::ERR_INCOMPLETE_CHUNKED_ENCODING`
And curl: `curl: (18) transfer closed with 1 bytes remaining to read`

I tried quite a few ways of changing finalhandler to not destroy the socket, or wait for some stream events, but can't figure it out.

Any ideas, even for workarounds, are greatly appreciated :)

great libs

### DIFF
--- a/test/session.js
+++ b/test/session.js
@@ -256,6 +256,82 @@ describe('session()', function(){
   })
 
   describe('when response ended', function () {
+    it('should not remove the last byte', function (done) {
+      var store = new session.MemoryStore();
+      // async
+      store.touch = function (sid, sess, fn) {
+        setTimeout(function () {
+          fn()
+        }, 5 * 1000)
+      };
+      var app = express()
+        .use(session({
+          store: store,
+          secret: 'keyboard cat',
+          resave: false
+        }))
+
+      var msg = 'I am the message. Don\'t strip my last byte'
+      app.get('/', function (req, res, next) {
+        // The following instrumentation is useful to debug
+        // It shows how _writeRaw will be called after
+        // res.connection is .destroyed()
+        // res.end = (function (og) {
+        //   return function () {
+        //     console.log("begin res.end")
+        //     var result = og.apply(this, arguments);
+        //     console.log('end res.end')
+        //     return result;
+        //   }
+        // }(res.end))
+        // res.connection.destroy = (function (og) {
+        //   return function () {
+        //     console.log("destroy")
+        //     return og.apply(this, arguments);
+        //   }
+        // }(res.connection.destroy))
+        // res._writeRaw = (function (og) {
+        //   return function (chunk) {
+        //     console.log("_writeRaw", chunk && chunk.toString)
+        //     return og.apply(this, arguments);
+        //   }
+        // }(res._writeRaw))
+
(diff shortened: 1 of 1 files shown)


dougwilson:
Hi! The bug is in your error middleware. You cannot call next() after you sent your response. Remember the Express rule: no next() after res.send(). You need to remove the invalid call to next(err) in your error handler that is already writing a response.

gobengo:
I removed that line and the test still fails.

 > Remember the Express rule: no next() after res.send().

Where would I remember that from? There are no mentions of anything like that on the README or API site. http://expressjs.com/en/4x/api.html

dougwilson:
Hi @gobengo , the referenced link does not document how to write middleware; the calling of `next()` is part of middleware writing, so I wouldn't expect it to be on the API page. You can read our middleware pages at http://expressjs.com/en/guide/writing-middleware.html and http://expressjs.com/en/guide/error-handling.html

As for your test case, the code is just not even a valid test case with that issue. You even explained the bug in our own code. If you are seeing your code enter into `finalhandler`, then you should never expect any response you tried to send to ever come though. The act of the code getting into `finalhandler` means you didn't properly handle the request and `finalhandler` is going into "last report mode" to try and close down the response. It does this in the following ways: if you didn't pass an `err` argument, it tries to send a 404, otherwise it tries to send a 500. If a response was already started and it gets into `finalhandler`, it will destroy the socket, because at that time, the assumption is if you are calling into `finalhandler`, you may not be writing any longer, so there is nothing left to do but just kill the socket to keep the client from waiting for bytes that may never come.

Your test cases contains multiple bugs you wrote into the code. Just like anything, if you are writing buggy code, of course it can result in bugs, that does not make it the fault of the module. If you're still not sure what specifically, please feel free to paste into this pull request code that I can just copy-and-paste directly into a file and run. I can then look at it and mark up as a response where the bugs are in the code.

The reason your test case in this pull request still fails after you removed the bad `next(err)` call, is because your line `res.get('content-length') !== Buffer.byteLength(res.text)` is invalid: you are strictly comparing a string to a number, so of course they will literally never match, even if they represent the same value. You either need to compare two numbers, or don't use a strict compare. For example, the following two conditions are valid: res.get('content-length') != Buffer.byteLength(res.text)` or `Number(res.get('content-length')) !== Buffer.byteLength(res.text)`.

gobengo:
Thanks

dougwilson:
No problem! If you still think there is a bug, please feel free to comment here more (I get emails of all comments) or even open a new issue.

PR #260, opened 2016-01-22
Update README.md

thepont:
Cookies are sent per hostname irrespective of port ( http://stackoverflow.com/questions/1612177/are-http-cookies-port-specific ) the documentation suggests that they are unique per hostname and port combination although this is not the case.

In that respect it is actually irrelevant if the servers are running on the same physical host, but only if the the hostname is the same for accessing the two applications.

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,7 +70,7 @@ request).
 
 The default value is `'connect.sid'`.
 
-**Note** if you have multiple apps running on the same host (hostname + port),
+**Note** if you have multiple apps running on the same hostname,
 then you need to separate the session cookies from each other. The simplest
 method is to simply set different `name`s per app.
 


gabeio:
I believe that is actually agreeing with you, except that they are talking about the same host *and* port for multiple *different* apps i.e.: behind a reverse proxy or even built into the same script so that the different sessions need to have different namespaces...

thepont:
But the port here is irrelevant, a web server on port 80 will be sent the same cookies as a web server running on 81 assuming they are on the same host.

So in that scenario you listed above you would get the same cookies, but also if you're running multiple http servers on the same host exposed with different ports.

gabeio:
> But the port here is irrelevant
Untrue, what I just said was that if you have multiple different domain names running on the same host and port you need to be able to separate them using the `name` which is in essence a namespace.

> So in that scenario you listed above you would get the same cookies, but also if you're running multiple http servers on the same host exposed with different ports.
If I got the same cookies for two different websites... that would be amazing.

gabeio:
Perhaps a compromise would be to make it: `if you have multiple apps running on the same port using different domains`? satisfying that we are taking about a single port not hostname+port ...

thepont:
I don't think you understand what I'm attempting to describe.

If you run two webservers on localhost one on 3333 and one on 3334 you will get the cookies from 3333 on the app running on 3334 and visa versa.

The port does not matter at _all_ and hence shouldn't be mentioned in the context.

gabeio:
Sorry, I do understand that the ports don't matter to the cookies. example of what you are saying:
```
HOST A:
  website.com:8081 > website.com:{any}
  website.com:8080 > website.com:{any}
```

I just think I was over thinking this.

thepont:
Yes.. Both 8081 and 8080 will get the same cookies.

dougwilson:
Hi @thepont, thank you for the pull request & sorry for the delay in merging it. I adding a little extra to it to help clarify what a hostname is to people. Let me know if this is OK or not :)

thepont:
Looks good, does a good job in clarifying how it will act.

PR #265, opened 2016-01-29
Add connect-dynamodb as a compatible session store

goyney:


dougwilson:
Thank you! Merged into the readme :)

PR #282, opened 2016-03-02
Defining built-in properties as non-enumerable.

gfoust:
I would like to be able to iterate over all data in a session, like so:

`for (var key in req.session) {
  console.log(key + '=' + req.session[key]);
}`

Currently this is complicated by the fact that the session object contains non-data properties such as `save`, `touch`, and `cookie`.  This patch simply defines those properties as non-enumerable using `Object.defineProperty` so that the above iteration technique works correctly.

### DIFF
--- a/session/session.js
+++ b/session/session.js
@@ -24,7 +24,6 @@ module.exports = Session;
 function Session(req, data) {
   Object.defineProperty(this, 'req', { value: req });
   Object.defineProperty(this, 'id', { value: req.sessionID });
-  Object.defineProperty(this, 'cookie', { value: undefined, writable: true })
 
   if (typeof data === 'object' && data !== null) {
     // merge data into this, ignoring prototype properties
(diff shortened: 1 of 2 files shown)


dougwilson:
Thanks! It looks like this change caused tests to fail if you can take a look, please :)

dougwilson:
In addition for taking a look into the failing tests this change caused, please also add at least one test that verifies the changes you made are working, otherwise your changes are likely to be un-done by another person and no one will noticed it's a problem since nothing will fail.

gfoust:
Will do, thanks.

gfoust:
I see now that MemoryStore depends on "cookie" being enumerable (which means other stores probably do as well).  I reverted that change; now only methods are non-enumerable.

I added a test case for the change.

dougwilson:
Awesome! Can you make those properties configurable, so users can redefine them, if there were doing that for some reason previously?

gfoust:
I made them configurable.  I'm not sure if this was what was desired, but I also made them writable, since they were writable previously.  Thus, the only change introduced by this patch is making them non-enumerable.

dougwilson:
Thank you so much! Yea, I'm not sure that they need to be configurable, but I just don't quite know what people are doing out there, and figured that getting this in a minor version that would at least be one less possible break :)

Issue #276, opened 2016-02-20
cookie.expire should be modified to match cookie.maxAge


QianJin2013:
Hi, i found when access server at second time, cookie.expire will not be modified even if i set maxAge and rolling:true.

Here is my test code. 

```js
var express = require('express');
var path = require('path');
var cookieParser = require('cookie-parser');
var bodyParser = require('body-parser');
var session = require('express-session');

var app = express();

// Use the session middleware
app.use(session({
  secret: 'keyboard cat',
  rolling: true,
  saveUninitialized: true,
  resave: true,
  cookie: { maxAge: 10000 }}))

// Access the session as req.session
app.get('/', function(req, res, next) {
  var sess = req.session
  if (sess.views) {
    sess.views++
    res.setHeader('Content-Type', 'text/html')
    res.write('<p>views: ' + sess.views + '</p>')
    res.write('<p>expires in: ' + (sess.cookie.maxAge / 1000) + 's</p>')
    res.end()
  } else {
    sess.views = 1
    res.end('welcome to the session demo. refresh!')
  }
});
```

In above example, i set **maxAge** to **10** seconds, and **rolling:true**, and saveUninitialized:true. Other settings seems have no effect to this problem.

I run this code, then use following curl command line to test and see verbose output:
```
$ curl --verbose --cookie cookie --cookie-jar cookie http://localhost:3000
* Rebuilt URL to: http://localhost:3000/
*   Trying 127.0.0.1...
* Connected to localhost (127.0.0.1) port 3000 (#0)
> GET / HTTP/1.1
> Host: localhost:3000
> User-Agent: curl/7.43.0
> Accept: */*
> 
< HTTP/1.1 200 OK
< X-Powered-By: Express
* Replaced cookie connect.sid="s%3AGRecnYYVv0WP0plY8qrIvV4EHG82Ts8h.aGG7rL0Fdb4fiXS0zN3u5eC9h%2BY8adZgZxzpyFZWFNs" for domain localhost, path /, expire 1455991539
< set-cookie: connect.sid=s%3AGRecnYYVv0WP0plY8qrIvV4EHG82Ts8h.aGG7rL0Fdb4fiXS0zN3u5eC9h%2BY8adZgZxzpyFZWFNs; Path=/; Expires=Sat, 20 Feb 2016 18:05:39 GMT; HttpOnly
< Date: Sat, 20 Feb 2016 18:05:29 GMT
< Connection: keep-alive
< Transfer-Encoding: chunked
< 
[...]

PR #296, opened 2016-03-29
Ensure the session is touched before the cookie is set

sboehler:


sboehler:
I initially struggled quite hard to find out what's happening in the patched _end method and found the separation of commit activities into things that happen 'on header' and in _end a bit arbitrary. Initially, to fix #276, I had the (simple) suggestion to touch the session twice, one 'on headers' and once in '_end' - admittedly not very elegant.

So I had another go and unified all session-handling code in 'on headers' and '_end' into one clean method. I believe my PR improves the code for a number of reasons:

- All session-related activities are cleanly isolated in commitSession
- The _end override is much simpler and readable
- it does not matter anymore whether the  'onHeaders' listener or 'end' is called first - the execution order is always the same (fixes #276)
- it is a lot more clear whether the session code is async or sync
- setting the cookie can be related to the other session commit activities, so future work on the code should be simpler and less error prone

All tests pass, including the new one for #276 , however I saw coverage decreased by 0.001%, not sure how to fix this. I am happy to get your comments and suggestions.

dougwilson:
Thanks @sboehler! The change is now very big, so it will take quite some time before I can review and accept it now, compared to the first PR, so please bear with me for the length of time it will take. The main reason there was a split is because of course by the time `_end` is called, you cannot set any headers, because they were already sent, so the logic that had to do with setting `Set-Cookie` header was there and saving the session was in the other location. In addition, the API for headers is a sync APi so it can only perform sync operations while the writing & end are async and can perform async operations like saving to the store. I hope that helps understand the split.

dougwilson:
In addition, and I don't think it's in the tests, your change *must* support the altering of `res.session` between writing the header and ending the response, or we cannot accept until the next major version. An example you may want to test if you haven't already:

```js
app.use(function (req, res) {
  req.session.foo = 'bar';
  res.write('<html>');
  setTimeout(function () {
    req.session.foo = 'baz';
    res.end('</html');
  }, 1000);
});
```

Should result in `foo = baz` in the data store, not `foo = bar` for compatibility with the current implementation.

dougwilson:
I also find your code much harder to read, particularly because it's altering `_callback` and there is a lack of whitespace, and I'm not inclined to accept the new rewrite just in general. I appreciate the effort, but the code needs to make sense and match the style of those of us who are maintaining it.

sboehler:
Hi @dougwilson , thanks for the quick feedback.

- Your example will likely break with my patch, as part of my idea is to start saving the session already 'on headers'. I wasn't aware of the requirement but I see why it is a breaking change. Also, it explains why you only start saving the session on _end, that's one thing I was confused about.

- White space could be fixed, no problem. As for _callback, the idea is to separate the _end logic from the session commit - I needed some way to transfer context from _end to the outside. While _end is being executed, we discover the task at hand and hence the context and the task at hand might change, therefore the reassignment.

The easy fix for #276 (which makes the 'rolling' option useless in practice in my use case) could then the initial proposal, touching the session twice (or some way via a global which remembers whether the session has been touched). Could you live with that? My principal ambition is that the behavior of rolling is restored, the other refactoring has been an insightful and fun exercise.

Thanks for your work with express, I appreciate it a lot.

PR #311, opened 2016-05-11
Added auto sub domain feature for cookies

rbjarnason:
We run one node.js server that serves multiple domains and we wanted to have the cookies shared across subdomains. The usual way to do this is to set domain: '.domain.com' in the options but this is not possible as we are serving multiple domains hence this simple addition.

### DIFF
--- a/index.js
+++ b/index.js
@@ -27,6 +27,7 @@ var Session = require('./session/session')
   , Cookie = require('./session/cookie')
   , Store = require('./session/store')
 
+var parseDomain = require('parse-domain');
 // environment
 
 var env = process.env.NODE_ENV;
@@ -145,6 +146,13 @@ function session(options){
 
     if (cookieOptions.secure === 'auto') {
       req.session.cookie.secure = issecure(req, trustProxy);
+    }
+
+    if (cookieOptions.autoSubDomain) {
+      var parsedDomain = parseDomain(req.headers.host);
+      if (parsedDomain && parsedDomain.domain) {
+        req.session.cookie.domain = '.' + parsedDomain.domain + '.' + parsedDomain.tld;
+      }
     }
   };
 
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "express-session",
-  "version": "1.13.0",
+  "version": "1.13.1",
   "description": "Simple session middleware for Express",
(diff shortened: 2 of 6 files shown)


package.json
```diff
@@ -18,7 +18,8 @@
     "on-headers": "~1.0.1",
     "parseurl": "~1.3.1",
     "uid-safe": "~2.1.0",
-    "utils-merge": "1.0.0"
+    "utils-merge": "1.0.0",
+    "parse-domain": "^0.2.1"
```
dougwilson: Please use an exact version and order this to be alphabetical (otherwise `npm` operations will reorganize it, creating unnecessary churn).

index.js
```diff
@@ -27,6 +27,7 @@ var Session = require('./session/session')
   , Cookie = require('./session/cookie')
   , Store = require('./session/store')
 
+var parseDomain = require('parse-domain');
```
dougwilson: If you can, please move this up with the other requires (probably right before the `parseUrl` one).

index.js
```diff
@@ -146,6 +147,13 @@ function session(options){
     if (cookieOptions.secure === 'auto') {
       req.session.cookie.secure = issecure(req, trustProxy);
     }
+
+    if (cookieOptions.autoSubDomain) {
+      var parsedDomain = parseDomain(req.headers.host);
```
dougwilson: Instead of only using `req.headers.host` directly, we need to actually hook into the mechanisms of Express.js such that it uses the settings and parsing that is in there so people who use "trust proxy" and the like can have the correct domain when using a reverse proxy.

index.js
```diff
@@ -146,6 +147,13 @@ function session(options){
...
+      if (parsedDomain && parsedDomain.domain) {
+        req.session.cookie.domain = '.' + parsedDomain.domain + '.' + parsedDomain.tld;
```
dougwilson: I'm curious here: why is the option called `autoSubDomain`, but yet this just leaves the `parsedDomain.subdomain` out of the picture? Perhaps I'm just not understanding what you're trying to do here. If you add the documentation for your feature to the README, I can read that and see if I understand :)

dougwilson:
Thanks for the pull request! On top of the comments I made inline, you also need to add documentation to the README for this feature as well as add unit tests that cover all new branches in code you've created here.

PR #331, opened 2016-07-19
Add missing sameSite property on cookie data object.

pazguille:
The sameSite property is not working. I just added the sameSite property on cookie data object.


gingermusketeer:
Was about to submit a PR with the same change. This fixes the issue for me as well.

dougwilson:
Thanks!

PR #332, opened 2016-07-19
Adding Medea session store

BenjaminVadant:
Hi,

Based on the level-session-store of Scripto, I've adapted a medea-session-store. 
The idea behind it is that it is fully functional on Android (as medeadown works easily on android but leveldown not).

Thank you for adding it.


dougwilson:
Awesome, thanks!

PR #336, opened 2016-07-28
Update README.md

nmhnmh:
Update README.md
- Add Usage code at the beginning of the doc
- Add separate configuration options section
- Adjust configuration section layout, to make nested cookie options like path, domain, maxAge looks obvious

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,49 +12,53 @@
 $ npm install express-session
 ```
 
-## API
-
+## Usage
 ```js
+var express = require('express')
 var session = require('express-session')
+var MongoStore = require('connect-mongo')(session);
+
+var app = express()
+
+app.use(session({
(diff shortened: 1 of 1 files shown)


dougwilson:
Thanks for the pull request! But, I don't think we want to accept it, mainly because it deviates from the standard formatting used across all the other repositories in this organization. If the proposal is to organize all the expressjs READMEs differently, we may want to bring it up in https://github.com/expressjs/discussions/issues first, discuss, and understand what the structure changes are, and how they help.

The usage section is also showcasing MongoDB, which is a very subjective example. Ideally, the example should be neutral, as otherwise it is not very useful for people not using MongoDB and being at the top leads to the suggestion that this module is somehow related to MongoDB and not immediately obvious that it can work with any database.

PR #337, opened 2016-08-01
Implement custom cookie signature

roblabla:
This is an updated version of #98, that applies properly on current master.

### DIFF
--- a/README.md
+++ b/README.md
@@ -276,6 +276,13 @@ The default value is `'keep'`.
   - `'keep'` The session in the store will be kept, but modifications made during
     the request are ignored and not saved.
 
+##### signature
+
+Pass in your own cookie signing object/module here that implements the
+`sign(value, secret)` and `unsign(value, secret)` functions.
+
+The default value is the `node-cookie-signature` module.
+
 ### req.session
 
 To store or access session data, simply use the request property `req.session`,
--- a/README.md
+++ b/README.md
@@ -281,7 +281,7 @@ The default value is `'keep'`.
 Pass in your own cookie signing object/module here that implements the
 `sign(value, secret)` and `unsign(value, secret)` functions.
 
-The default value is the `node-cookie-signature` module.
+The default value is the [`cookie-signature`](https://github.com/tj/node-cookie-signature) module.
 
 ### req.session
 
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -602,8 +606,8 @@ function issecure(req, trustProxy) {
  * @private
  */
 
-function setcookie(res, name, val, secret, options) {
-  var signed = 's:' + signature.sign(val, secret);
```
dougwilson: This change is not backwards-compatible, just as a FYI. It would also make the cookies emitted from here unreadable by other modules trying to decode them without them changing. What is the purpose of this change?

dougwilson:
Looks like the PR is marked as failed because there are not enough tests to cover all the changes made in the PR. Please update the tests to try to cover all added code paths :)

README.md
```diff
@@ -276,6 +276,13 @@ The default value is `'keep'`.
   - `'keep'` The session in the store will be kept, but modifications made during
     the request are ignored and not saved.
 
+##### signature
+
+Pass in your own cookie signing object/module here that implements the
+`sign(value, secret)` and `unsign(value, secret)` functions.
+
+The default value is the `node-cookie-signature` module.
```
dougwilson: Typo here, as this module uses the `cookie-signature` module, not `node-cookie-signature`.

dougwilson: We could probably include a link to the npm page for the module as well.

index.js
```diff
@@ -78,6 +78,8 @@ var defer = typeof setImmediate === 'function'
  * @param {Boolean} [options.saveUninitialized] Save uninitialized sessions to the store
  * @param {String|Array} [options.secret] Secret for signing session ID
  * @param {Object} [options.store=MemoryStore] Session store
+ * @param {Object} [options.signature] Object that has the same API as node-cookie-signature
```
dougwilson: Typo: node-cookie-signature -> cookie-signature

Also, I'm not sure we should define our API as "being the same as another module". Perhaps just spell out what we want and just say that `cookie-signature` is what is following our API, instead of the other way around.

index.js
roblabla: My purpose here was that I need to have full control over the cookie's content. (My end-game is to make express-session compatible with PHP's session management). That means having a "pass-through" signature function (AKA no signature at all).

I have a patch that should (in theory) turn back compatibility while still allowing me to achieve my goals.

roblabla:
Done with all your suggestions :). I think it should be compatible with previous versions now. The `s:` is only appended in the default signature signature. In my view, it's the job of the signature to append it or not. This allows express-session to be compatible with php's PHPSESSID without much trouble.

roblabla:
Bump. Is there anything missing from this PR ?

PR #346, opened 2016-08-24
Update README.md

jackspaniel:
Add link to Hazelcast store in available session stores


jackspaniel:
Updating README with link to Hazelcast session store

PR #369, opened 2016-10-08
fix: 'setcookie' - 'cookie.serialize' can't return Array, its result is always String

olegabyst:


dougwilson:
I'm not 100% sure what is being fixed. Can you please update the PR with the following?

1. Please elaborate on what is broken and what the fix is trying to do.
2. Please take a look at all the failing tests from this change and get the existing tests passing.
3. Please add a new test that does not pass before your fix but passes after you fix. This makes sure we have a test for the bug you are fixing so we don't accidentally break it again later.

Thanks!

OlegTsyba:
since "cookie.serialize" always return String, function "setcookie" contain unnecessary check.

dougwilson:
Ah, I see what you are saying now. So currently that code is just a copy-paste we are sharing between expressjs modules that is how you append to a header field; it is just the template
```js
function appendHeader (name, data) {
  var prev = res.getHeader(name) || [];
  var header = Array.isArray(prev) ? prev.concat(data)
    : Array.isArray(data) ? [prev].concat(data)
    : [prev, data];

  res.setHeader(name, header)
}
```

But simply inline. I understand that `cookie.serialize` will only return a string and not an array, but if we change this, it's a lot harder to keep these snippets up-to-date between all the modules.

Does it at least provide some benefit for the cost of increased maintenance? Perhaps some kind of measurable performance boost?

OlegTsyba:
I didn't measure whether it gives perceptible performance boost, I think my commit makes code simpler and eliminates unnecessary runtime check. I was not aware that it's some copy-pasted pattern.

dougwilson:
It technically increases the branch test coverage, since that branch was not covered by tests (and couldn't be). I merged it in for the next release, thanks!

PR #370, opened 2016-10-12
Adding the MarkLogic Server session store connect-ml

bluetorch:
Added a new session store for MarkLogic Server called connect-ml


dougwilson:
Thanks!

Issue #2, opened 2014-02-16
Max age resetting after response


jonathanong:
Is there a reason the max age is reset with each response?
I need to be able to see if half of TTL has elapsed. The cookies expiration is being increased but the session still expires at the original max age.

``` JavaScript
   // proxy end() to commit the session
    var end = res.end;
    res.end = function(data, encoding){
      res.end = end;
      if (!req.session) return res.end(data, encoding);
      debug('saving');
      req.session.resetMaxAge();
      req.session.save(function(err){
        if (err) console.error(err.stack);
        debug('saved');
        res.end(data, encoding);
      });
    };
```

timaschew:
> Fixing this would require a major version bump, sadly :(
> dougwilson added this to the 2.0.0 milestone on 5 Jan 2015

express 4 is out now, what's the reason why the bug is not fixed yet?

dougwilson:
@timaschew , I'm unsure what the major version of this module has to do with the version of Express.

HarshithaKP:
@jonathanong Just wondering this is not a bug. Once client requests  and is being active, to keep the session interactive, `maxAge` of cookie is been refreshed with each response as cookie expiry disconnects client from session with ongoing interactions.


cc @dougwilson

dougwilson:
Hi @HarshithaKP I have never heard of the issue you are describing. Can you provide code and steps to reproduce the issue you are referring to?

HarshithaKP:
@dougwilson Sorry that assumption was a mistake, I will go through the issue again and get back to you.

davidmashe:
This issue is not prioritized and has not had meaningful activity for some time. It can be re-opened when a PR is raised, if needed.

PR #376, opened 2016-10-26
fixed "Max age resetting after response #2" open issue

iggirex:
https://github.com/expressjs/session/issues/2

This issue did not allow users to specify a boolean for rolling (property when setting cookie). After this fix, rolling can be set to boolean and works.

### DIFF
--- a/index.js
+++ b/index.js
@@ -428,7 +428,7 @@ function session(options) {
 
       return cookieId != req.sessionID
         ? saveUninitializedSession || isModified(req.session)
-        : rollingSessions && req.session.cookie.expires && isModified(req.session);
+        : rollingSessions || req.session.cookie.expires && isModified(req.session);
     }
 
     // generate a session if the browser doesn't send a sessionID
--- a/index.js
+++ b/index.js
@@ -428,7 +428,7 @@ function session(options) {
 
       return cookieId != req.sessionID
         ? saveUninitializedSession || isModified(req.session)
-        : rollingSessions || req.session.cookie.expires != null && isModified(req.session);
+        : rollingSessions && req.session.cookie.expires && isModified(req.session);
     }
 
     // generate a session if the browser doesn't send a sessionID
--- a/session/memory.js
+++ b/session/memory.js
@@ -149,6 +149,7 @@ MemoryStore.prototype.touch = function touch(sessionId, session, callback) {
 
   if (currentSession) {
     // update expiration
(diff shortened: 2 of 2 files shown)


dougwilson:
Hi @iggirex, thank you for the pull request! It looks like you didn't add any new tests for your change and the existing tests are no longer passing, which indicates there is either a mistake in the PR, or this is a major version bump change. Can you elaborate more here on exactly what the specific changes in behavior this PR is making?

iggirex:
Sure thing!

My collaborator and I were going through figuring out how to make the rolling property on cookies work, because if it's set to true or false, the behavior is the same: the max age of the cookie gets reset everytime a response is sent.

  function shouldSetCookie(req) {
      // cannot set cookie without a session ID
      if (typeof req.sessionID !== 'string') {
        return false;
      }

      return cookieId != req.sessionID
        ? saveUninitializedSession || isModified(req.session)
        : rollingSessions || req.session.cookie.expires && req.session.cookie.expires != null && isModified(req.session);
    }

We changed the underlined part of the code so it does not check that the cookie.expires isn't null. The way we found this to work was more of trial and error than understanding the nuances of what is happening here. The tests were failing because I turned the inital || after rollingSessions to an &&. Once it's changed back, all the tests seem to be passing. I'll send another PR.

thanks for the quick response!
Iggi

dougwilson:
Ah, I see @iggirex. Please add a test here as well that tests the new behavior you have added.

iggirex:
Sure thing!

My collaborator and I were going through figuring out how to make the
rolling property on cookies work, because if it's set to true or false, the
behavior is the same: the max age of the cookie gets reset everytime the a
response is sent.

  function shouldSetCookie(req) {
      // cannot set cookie without a session ID
      if (typeof req.sessionID !== 'string') {
        return false;
      }

      return cookieId != req.sessionID
        ? saveUninitializedSession || isModified(req.session)
        : rollingSessions || req.session.cookie.expires &&
*req.session.cookie.expires
!= null &&* isModified(req.session);
    }

We changed the underlined part of the code so it does not check that the
cookie.expires isn't null. The way we found this to work was more of trial
and error than understanding the nuances of what is happening here. The
tests were failing because I turned the inital || after rollingSessions to
an &&. Once it's changed back, all the tests seem to be passing. I'll send
another PR.

thanks for the quick response!
Iggi

dougwilson:
I'm going to close this PR since there hasn't been any changes from the multiple requests to add tests.

Issue #281, opened 2016-03-02
unable to use secure cookies


fibo:
I am trying to set secure cookies, it does not work at first try.

I am using

```js
 secret: sessionSecret,
      resave: false,
      saveUninitialized: true,
      cookie: {
        secure: true,
        httpOnly: true,
        domain: 'beintoo.net',
        expires: expiryDate
      }
```

I found function `issecure` relying on `req.connection.encrypted` [here](https://github.com/expressjs/session/blob/v1.13.1/index.js#L543) but it seems not supported anymore, see https://github.com/expressjs/express/issues/1864

Any hint?

dougwilson:
Hi @fibo, is your Express server even listening on a SSL socket (`https.createServer(app, options)`) or are you terminating your SSL with a reverse proxy in front of your Node.js app?

dougwilson:
Also, if you haven't already, please take a read through our documentation on this at https://github.com/expressjs/session#cookie-options and see if perhaps you just forgot to set some of those Express configurations or something.

fibo:
Hi @dougwilson , I am using nginx to handle SSL. I had a look at the documentation but I could not found a solution. I found the line of code linked above cause I enabled debug with `export DEBUG=express-session` and saw the *not secured* message.

I have just tryed *cookie-session* Package and iwth the `secureProxy` option enabled worked almost out of the box.

Please let me know if I can contribute back with some additional info or test, otherwise close the issue.

dougwilson:
Hi @fibo, what version of Express are you using? Have you tried adding the `app.set('trust proxy', 1)` to your code like in our example?

fibo:
Yes, I tryed to set "trust proxy" as required in the documentation.

Unfortunately I am working on a private project right now so it is more difficult to show you. By the way, did you checked the line of code linked above? It refers to `req.connection.encrypted` which seems to be not valid anymore.

I am using Node v4.2.2, here are my deps:

```
  "dependencies": {
    "body-parser": "^1.14.1",
    "connect-ensure-login": "^0.1.1",
    "cookie-session": "^2.0.0-alpha.1",
    "debug": "^2.2.0",
    "express": "^4.13.3",
    "express-session": "^1.12.1",
    "jade": "^1.11.0",
    "mdconf": "0.0.2",
    "passport": "^0.3.2",
    "passport-local": "^1.0.0",
    "pg": "^4.4.3",
    "pg-promise": "^3.1.4",
    "pm2": "^1.0.0",
[...]

PR #381, opened 2016-10-29
Reword cookie.secure documentation

au:
I'd like to double check on whether or not `proxy: true` is necessary, but this new documentation should fix some common problems people have when using reverse proxy / unwittingly enable secure cookies.

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,25 +112,35 @@ Specifies the `boolean` value for the `Secure` `Set-Cookie` attribute. When trut
 the `Secure` attribute is set, otherwise it is not. By default, the `Secure`
 attribute is not set.
 
-**Note** be careful when setting this to `true`, as compliant clients will not send
-the cookie back to the server in the future if the browser does not have an HTTPS
-connection.
+Securing cookies is **recommended**. By connecting over HTTP, both the 
+user and session can be easily exposed. To properly configure secure cookies, 
+some configuration is necessary:
 
-Please note that `secure: true` is a **recommended** option. However, it requires
-an https-enabled website, i.e., HTTPS is necessary for secure cookies. If `secure`
-is set, and you access your site over HTTP, the cookie will not be set. If you
-have your node.js behind a proxy and are using `secure: true`, you need to set
-"trust proxy" in express:
+- Secure cookies must be served through **HTTPS**. Most clients will not send 
+secure-marked cookies over HTTP. 
+- If node is behind a proxy, Express *and* Session.js must be told to trust the proxy.
 
 ```js
 var app = express()
-app.set('trust proxy', 1) // trust first proxy
+app.set('trust proxy', 1)
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -112,27 +112,37 @@ Specifies the `boolean` value for the `Secure` `Set-Cookie` attribute. When trut
...
-is set, and you access your site over HTTP, the cookie will not be set. If you
-have your node.js behind a proxy and are using `secure: true`, you need to set
-"trust proxy" in express:
+- Secure cookies must be served through **HTTPS**. Most clients will not send 
+secure-marked cookies over HTTP. 
+- If node is behind a proxy, Express *and* Session.js must be told to trust the proxy.
 
 ```js
 var app = express()
-app.set('trust proxy', 1) // trust first proxy
```
dougwilson: Why remove the comment `trust first proxy` ?

README.md
```diff
@@ -112,27 +112,37 @@ Specifies the `boolean` value for the `Secure` `Set-Cookie` attribute. When trut
...
+app.set('trust proxy', 1)
+
 app.use(session({
-  secret: 'keyboard cat',
-  resave: false,
-  saveUninitialized: true,
+  ...
```
dougwilson: We typically like these to be valid JS, though the `...` is not.

README.md
```diff
@@ -112,27 +112,37 @@ Specifies the `boolean` value for the `Secure` `Set-Cookie` attribute. When trut
...
+  proxy: true,
   cookie: { secure: true }
 }))
 ```
 
+- If using nginx as a proxy, nginx must be configured to send its current protocol,
```
dougwilson: Please remove the whole nginx section; if you get it added to expressjs.com we'll link there; we don't want to becomes the main source for how Express works, especially since this module works with non-Express frameworks.

README.md
```diff
@@ -144,8 +154,8 @@ var sess = {
 }
 
 if (app.get('env') === 'production') {
-  app.set('trust proxy', 1) // trust first proxy
```
dougwilson: Why remove these comments?

au:
Oh; the main purpose of the PR was to add documentation for errors devs had using nginx as a reverse proxy. If that isn't a goal of yours, the pr may be closed.

dougwilson:
Hi @au , sorry, I guess the other issues were confusing; we don't want to document nginx, apache, lighttpd, etc. here, because this module doesn't do anything, and this works with more than Express.js. We want those docs to be located on expressjs.com along with their trust proxy docs and we can link to that as the source of truth, since they need those docs just like we do, so there is no point of having to maintain two copies when we can just link. The others parts of your PR are good, though.

dougwilson:
https://github.com/expressjs/session/issues/281#issuecomment-200329301 is my recommendation.

PR #384, opened 2016-10-30
Wrap session.reload

### LINKED ISSUE #383: session.reload causes session resave at the end of the request
To avoid race conditions, whenever we change the session, we do `session.reload` & `session.save`.
We also use a custom synchronous session store, so session changes are atomic.
But I noticed that the session is saved again at the end of the request. This leads to the following sequence:
1. session.reload
2. session.save
3. session.save

Any parallel session changes between step 2 and 3 are overwritten by the save in step 3.
This leads again to race conditions.

It seems that session methods are not wrapped after `session.reload` and any further saves are not registered by _express-session_. So it thinks the session is not saved and saves it at the end of the request.

dotchev:
Fix #383

### DIFF
--- a/index.js
+++ b/index.js
@@ -364,7 +364,9 @@ function session(options) {
 
     // wrap session methods
     function wrapmethods(sess) {
+      debug('wrapping %s', sess.id);
       var _save = sess.save;
+      var _reload = sess.reload;
 
       function save() {
         debug('saving %s', this.id);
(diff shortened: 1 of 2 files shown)


dougwilson:
Thanks! When you get a chance, can you add at least one test that fails before your fix and passes after your fix? We don't want to break this in the future :)!

dotchev:
Thanks for the merge.
Could you release a new version so we can use the fix?

dougwilson:
I will release soon. Until then, just put the repo commit in your `package.json` directly.

PR #390, opened 2016-11-28
Remove buffer deprecated warning

citium:
crc 3.4.2/3 -- have replace buffer() with new Buffer(), hence remove the deprecation warning

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
   "dependencies": {
     "cookie": "0.3.1",
     "cookie-signature": "1.0.6",
-    "crc": "3.4.1",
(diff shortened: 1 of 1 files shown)


dougwilson:
Thanks, @citium ! It seems your description is misleading; crc version 3.4.1 removed the usage that caused a deprecation warning, which is why you don't see the warning emitted when our Travis CI is run. 3.4.2/3.4.3 just replaced `new Buffer` (which is only deprecated in docs, not actually in code).

citium:
@dougwilson, thanks I'll will try to articulate better next time.

PR #392, opened 2016-12-02
docs: update readme about cookie-parser dependency

kevinsimper:
There is some confusing about wheater using `express-session` eliminate the need for cookie-parser complete or this specific modules simply doesn't need it.

Just one example:
http://stackoverflow.com/questions/27961320/when-should-i-use-cookie-parser-with-express-session

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ Create a session middleware with the given `options`.
 Session data is stored server-side.
 
 **Note** Since version 1.5.0, the [`cookie-parser` middleware](https://www.npmjs.com/package/cookie-parser)
-no longer needs to be used for this module to work. This module now directly reads
+no longer needs to be used for this module to work, however other modules may still need it like [`csurf`](https://github.com/expressjs/csurf). This module now directly reads
 and writes cookies on `req`/`res`. Using `cookie-parser` may result in issues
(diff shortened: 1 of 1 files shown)


joewagner:
@kevinsimper Thanks for the PR
My thought is that if `csurf` stops requiring `cookie-parser` we have docs that are inaccurate.  Maybe just state something to the affect that `cookie-parser` isn't necessary for this module, but it won't conflict with this module if you need to use it for another reason.
Would that clarify things more?

dougwilson:
And on @joewagner 's comments, the `csurf` module is a weird choice to even use as an example, as it only sometimes needs `cookie-parser`, and that's only when you _not using it with this module_.

kevinsimper:
@joewagner Yes, I think that will clarify it!

@dougwilson The depency and order of these modules is steep, getting confused but I understand what you are saying.

gabeio:
@kevinsimper basically csurf will use this module if present... otherwise it will need a way to give the user a cookie to compare against the token sent back along with form content.

dougwilson:
Since there has been conversation about the issues in the PR and no changes made to the PR, I'm going to assume that the PR is mostly stalled and close it for now. Let me know if you want to make changes and re-open it (or make a new PR).

PR #393, opened 2016-12-05
add warning content about saving data.

levin-du:
update readme.md. add warning content about  saving data.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,6 +39,8 @@ not designed for a production environment. It will leak memory under most
 conditions, does not scale past a single process, and is meant for debugging and
 developing.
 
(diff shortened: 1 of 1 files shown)


levin-du:
hi there, I found the solution after hours debug and read session source code. I thought It may saving others time by add the warning content. Many users ask questions on StackOverflow.com when they face the same situation. This may be worth.

dougwilson:
Thanks, but it's pretty much a duplicate of the existing docs:

> This method is automatically called at the end of the HTTP response if the session data has been altered (though this behavior can be altered with various options in the middleware constructor).

levin-du:
The existing docs does not address the problem out: Update session after response.send or response.end will not work.

PR #401, opened 2016-12-16
Allow user to set some special urls which be requested would not invoke .touch() function.

VampireDaniel:


dougwilson:
Hi @VampireDaniel I'm not sure how I feel about this addition, but at the very least, it doesn't seem very useful, since this won't work for anyone who has case-sensitive URLs (since URLs are case-sensitive by default in the specs, it's probably fairly common) and it breaks down as soon as a user adds a query string parameter to the URL, causing a mis-match.

VampireDaniel:
Hi @dougwilson ,Thanks for reply. This idea comes from my real case.
For some security reason we cant get session expiry date via http request nor from cookies in local.
but I need to do something referring the expiry date, and I think this is a way to realize my demand.

Yes, I have found urls with query string would cause some bugs, I will do more test and I will to use regular expression for matching.

gabeio:
aa816ce changes multiple lines of the package.json. If you are planning on leaving this pr open I'd suggest either removing/reverting that change or opening a new pr with a non-master branch of your fork.

joewagner:
@VampireDaniel It looks like you changed the readme to ask users to install a different package. Are you trying to rename this package on npm?

As an aside: Instead of adding complexity to this package, can you handle this logic in the routing part of your application?  maybe you could only apply this middleware to the urls you want touched?

VampireDaniel:
@all sorry I was so busy those days, I have renamed the package on npm, and I am improving the my code and adding tests, I'll close this request and re-pull again.

PR #418, opened 2017-01-23
Pass options to cookie.parse to avoid defaults

caub:
To improve perf most likely

https://github.com/jshttp/cookie/blob/master/index.js#L55-L57

### DIFF
--- a/index.js
+++ b/index.js
@@ -508,6 +508,7 @@ function generateSessionId(sess) {
  * @return {string}
  * @private
  */
+var cookieDefaultOptions = {decode: decodeURIComponent, encode: encodeURIComponent};
 
 function getcookie(req, name, secrets) {
   var header = req.headers.cookie;
@@ -516,7 +517,7 @@ function getcookie(req, name, secrets) {
 
   // read from cookie header
   if (header) {
-    var cookies = cookie.parse(header);
+    var cookies = cookie.parse(header, cookieDefaultOptions);
 
     raw = cookies[name];
 


dougwilson:
Can you show what the perf gain is? This is harder to maintain, so I would like to be sure the gain is worth it.

dougwilson:
Also the change breaks the jsdocs for the function.

caub:
```js
var cookie = require('cookie');

var cookieDefaultOptions = {decode: decodeURIComponent, encode: encodeURIComponent};

// console.time(1);console.timeEnd(1);
function test1(){
	console.time(1);
	for (var i=0; i<1e4; i++) {
		cookie.parse('foo=%1;bar=bar;HttpOnly;Secure');
		cookie.parse('foo=%1;bar=bar;foo=boo');
		cookie.parse('foo=%1;bar=bar');
		cookie.parse('foo="bar=123456789&name=Magic+Mouse"');
	}
	console.timeEnd(1);
}
function test2(){
	console.time(2);
	for (var i=0; i<1e4; i++) {
		cookie.parse('foo=%1;bar=bar;HttpOnly;Secure', cookieDefaultOptions);
		cookie.parse('foo=%1;bar=bar;foo=boo', cookieDefaultOptions);
		cookie.parse('foo=%1'+i+';bar=bar', cookieDefaultOptions);
		cookie.parse('foo="bar=123456789&name=Magic+Mouse"', cookieDefaultOptions);
	}
	console.timeEnd(2);
}


setImmediate(test2)
setImmediate(test1)
setImmediate(test2)
setImmediate(test1)
setImmediate(test2)
setImmediate(test1)

/*

2: 305.208ms
1: 346.909ms
2: 304.842ms
1: 376.245ms
2: 290.874ms
1: 371.629ms
[Finished in 2.3s]
*/

```

joewagner:
Why is there a `+i+` in line 21 for test 2, and not in test1?

When I remove that I get the following results
2: 258.397ms
1: 251.719ms
2: 233.395ms
1: 244.811ms
2: 235.622ms
1: 244.035ms

Which seems almost identical

caub:
sorry I corrected, it's hard to see the difference now

dougwilson:
So from the conversation above, it doesn't sound like this really does much, and the jsdoc breakage hasn't been fixed. I'm going to close this, but let me know if I misunderstood the resolution.

PR #419, opened 2017-01-27
feat: Option to allow regenerate() to preserve old session

mccleanp:
Add preserve option to session.regenerate() function for use cases when it is desirable to regenerate the session, but not destroy the old session immediately. For example when there could be multiple ajax requests inflight with the old session id.

Existing functionality is preserved if the option is omitted. Existing test was improved to cover session destuction and new test was added for the preserve case.

This was motivated by the desire to implement middleware to provide session renewal, as recommended by [OWASP](https://www.owasp.org/index.php/Session_Management_Cheat_Sheet#Renewal_Timeout)

### DIFF
--- a/session/store.js
+++ b/session/store.js
@@ -47,12 +47,31 @@ util.inherits(Store, EventEmitter)
  * @api public
  */
 
-Store.prototype.regenerate = function(req, fn){
+Store.prototype.regenerate = function regenerate (req, options, fn) {
+  var cb = fn
+  var opts = options || {}
   var self = this;
-  this.destroy(req.sessionID, function(err){
-    self.generate(req);
-    fn(err);
-  });
+
+  if (typeof options === 'function') {
+    cb = options
+    opts = {}
+  }
+
+  var destroy = opts.destroy !== undefined
+    ? opts.destroy
+    : true
+
+  if (destroy) {
+    this.destroy(req.sessionID, function (err) {
+      self.generate(req)
+      cb(err)
+    })
+  } else {
+    process.nextTick(function () {
+      self.generate(req)
+      cb(null)
+    })
+  }
 };
 
 /**
--- a/README.md
+++ b/README.md
@@ -304,15 +304,31 @@ app.get('/', function(req, res, next) {
 })
 ```
 
-#### Session.regenerate(callback)
+#### Session.regenerate([options], callback)
 
-To regenerate the session simply invoke the method. Once complete,
+Regenerates the session for the current request. Once complete,
 a new SID and `Session` instance will be initialized at `req.session`
 and the `callback` will be invoked.
 
+#### Options
+
+If the options argument is omitted, the options will take their default values.
+The following options are supported:
+
+##### destroy
+
+Specifies whether the existing session should be destroyed or not. Defaults
+to `true`, and the existing session will be destroyed. If this is not desirable,
+set to `false` to preserve the existing session.
+
 ```js
 req.session.regenerate(function(err) {
-  // will have a new session here
+  // existing session is destroyed and will have a new session here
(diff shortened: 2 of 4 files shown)


dougwilson (changes requested):
Nice, I think this is a great feature to add and make sense to me. I really only have two asks:

1. We'll need to get documentation added to the README.
2. I feel like the API is a little weird, and generally feel like APIs that take a Boolean as a random argument is hard to remember that a `true` means without looking it up and also it's even odder to have arguments appearing after a callback argument. I would think maybe the API be `req.session.regenerate({ preserve: true }, callback)`

mccleanp:
I have pushed another commit to address the comments from @dougwilson. Let me know if you want me to squash the commits.

session/store.js
```diff
@@ -47,12 +47,27 @@ util.inherits(Store, EventEmitter)
  * @api public
  */
 
-Store.prototype.regenerate = function(req, fn){
+Store.prototype.regenerate = function(req, options, fn){
   var self = this;
-  this.destroy(req.sessionID, function(err){
+  if (options instanceof Function) {
```
dougwilson: I would recommend using `typeof options === 'function'` instead of `instaceof`.

session/store.js
```diff
@@ -47,12 +47,27 @@ util.inherits(Store, EventEmitter)
...
+    fn = options;
+    options = {};
+  }
+  options = options || {};
```
dougwilson: Back in commit https://github.com/expressjs/session/commit/d95275bd5345b161585ec7c6de22bcd77b484a65 we tried to fix some deopts by avoiding argument reassignments. Please avoid argument reassignments in this code as well :)

session/store.js
```diff
@@ -47,12 +47,27 @@ util.inherits(Store, EventEmitter)
...
+  if ( destroy ) {
+    this.destroy(req.sessionID, function(err){
+      self.generate(req);
+      fn(err);
+    });
+  } else {
     self.generate(req);
-    fn(err);
-  });
+    fn(null);
```
dougwilson: Please make sure you don't "release Zalgo" by making an API sometimes sync, and sometimes async. See http://blog.izs.me/post/59142742143/designing-apis-for-asynchrony for a complete explanation. The API should be always one or the other.

dougwilson (changes requested):

dougwilson:
Hi @mccleanp thanks for updating the PR! What ever happened to the `preserve: true` above? I don't remember you mentioning that you would be implementing it a different way. The reason I ask is because usually in the JS sense, an option that defaults to `false` is easy to understand, since `null` and `undefined` is falsy, which people associate with default values. This implementation defaults the option to `true`.

mccleanp:
@dougwilson: I am happy to change it back to `preserve: true` if you prefer. I had selected `destroy: false` simply because the destroy terminology was already used, while `preserve` is introducing new terminology.

dougwilson:
Yea, if you have a better idea for the terminology, that's great, just the key constraint I'm trying to keep here is that the default value should be `false`, not `true`, for better mental mapping of `undefined` / `null`.

Flame2057:
Ok looks good let's merge it :)

Flame2057:
I implemented this feature as an express middleware. It's on npm here:
https://www.npmjs.com/package/express-session-fixation

dougwilson:
I just rebased the commits. I'm going to work on cleaning it up further to merge 👍

max8hine:
I currently have a problem is that the previous sessionStore got delete after run regenerate() function.

Not sure is it could be solved with this PR. but I think to let the old sessionStore expire itself by `cookie.maxAge` make more sense.

PR #324, opened 2016-06-15
WIP: changes to standardjs code style

gabeio:
There are still a few(actually a lot of the same) errors on the standard that I'm not understanding how to fix (90% are the test/session.js).

here's just a few:
- [x] `session/index.js:87:7: 'options' is already defined`
- [x] `session/index.js:162:59: Unexpected use of comma operator.`
- [x] `session/session/store.js:30:1: The '__proto__' property is deprecated.`
- [x] `session/test/session.js:17:1: 'describe' is not defined.`
- [x] `session/test/session.js:18:3: 'it' is not defined.`
- [x] `session/test/session.js:27:20: Expected consistent spacing`
- [x] `session/index.js:447: Unexpected mix of '||' and '&&'  no-mixed-operators`
- [ ] `session/test/session.js:2350: Unnecessary escape character: \.  no-useless-escape`

PR #422, opened 2017-02-03
Make variable type checks consistent

chris-kobrzak:
Whenever you use the `typeof` operator, you are guaranteed to be working with
strings, as this is what typeof returns. For this reason it is generally a good
idea to use the equality operator instead of the strict equality operator for
comparisons.

This is a minor change and it is consistent with other areas of the code where
the equality operator is already in use, e.g. [session/session.js](https://github.com/expressjs/session/blob/8e57b21bef0d53010c4d37f4beee3f0341f3eaa6/session/cookie.js#L72)

### DIFF
--- a/index.js
+++ b/index.js
@@ -61,7 +61,7 @@ var warning = 'Warning: connect.session() MemoryStore is not\n'
  */
 
 /* istanbul ignore next */
-var defer = typeof setImmediate === 'function'
+var defer = typeof setImmediate == 'function'
   ? setImmediate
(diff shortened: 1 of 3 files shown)


dougwilson:
Thanks for the PR, but we are trying to move to code to use StandardJS style (see other PR) and this is going to opposite direction.

dougwilson:
StandardJS PR is #324

chris-kobrzak:
Ah, makes sense, thanks for the heads-up, @dougwilson!

PR #441, opened 2017-03-11
Fixes sightly confusing versioning typo in README

peteratticusberg:
The latest version of express-session is 1.15.1, though the note concerning cookie-parser states that "since version 1.5.0" this dependency is no longer required.


dougwilson:
That is correct, it is no longer required as of version 1.5.0. I'm not really sure what is confusing about that... ?

dougwilson:
And here is the changelog entry for version 1.5.0 where we removed the requirement for cookie-parser: https://github.com/expressjs/session/blob/master/HISTORY.md#150--2014-06-19

PR #444, opened 2017-03-20
add option "expireSetCookie" to determine if set 'Expire' attribute of…

asdw741111:
… Set-Cookie in header or not.

add option "expireSetCookie" to determine if set 'Expire' attribute of Set-Cookie in header or not.
Session expire only keep in server,otherwise it will occur sessionid regenerate,because browser will not refresh cookie expire time.


dougwilson:
Closing since the requests changes were never made.

PR #447, opened 2017-03-22
Update documentation for Session.touch()

Ionaru:
Session.touch takes a callback as argument
The `.maxAge` property is only updated when `rolling` is `true`

Information from http://stackoverflow.com/questions/14464873/expressjs-session-expiring-despite-activity

### DIFF
--- a/README.md
+++ b/README.md
@@ -360,10 +360,9 @@ req.session.save(function(err) {
 })
 ```
 
-#### Session.touch()
+#### Session.touch(callback)
 
-Updates the `.maxAge` property. Typically this is
-not necessary to call, as the session middleware does this for you.
+Updates the `.maxAge` property. This is done automatically when `rolling` is `true`.
 
 ### req.session.id
 


dougwilson:
Hi @Ionaru I'm sorry if you are finding conflicting information on a SO post from 4 years ago, but things have changed in those four years. The README represents the current version of this module and both changes made in this PR are sadly not correct.

The `Session#touch` method does not take a callback; you can find the implementation of this method here: https://github.com/expressjs/session/blob/master/session/session.js#L47-L49

The max age of the cookie is already reset when the cookie is set, regardless if `rolling` is true or false.

Ionaru:
In that case the typing information for Express-session is horribly outdated as well.

Cookies in my application were not "touched" on every request when `rolling` is false, so I assumed it to be incorrect.

dougwilson:
> In that case the typing information for Express-session is horribly outdated as well.

I'm not familiar with the typings; I don't think that's something in this repository?

> Cookies in my application were not "touched" on every request when rolling is false, so I assumed it to be incorrect.

That's correct, cookies are not touched on every request when rolling is false; they are only touched when the session is modified. The update to the `Session#touch` description didn't really explain anything relating to this, because the max age of the cookie is already reset when the cookie is set, regardless if rolling is true or false.

dougwilson:
FWIW, the `Session#touch` method has never taken a callback in any version of this module, so no idea where that information would have come from.

Ionaru:
The information came from the [@types/express-session](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/express-session) package from DefinitelyTyped

Some more information on `Session#touch` would be useful in that case. In addition of saying the session middleware calls the function, explain when and in which situation it does that.

PR #458, opened 2017-05-02
docs: add express-session-cache-manager to the list of session stores

theogravity:
@dougwilson 

sorry about that


dougwilson:
No problem :) ! It did take me a while to figure out why I couldn't download the patch, but now I know 💯

PR #461, opened 2017-05-07
Ensure the session is saved before sending first chunk

andykog:
For now if asyncronous session saving takes little more then reading, you start dealing with the problem described in #74 The problem is also covered in updated tests.
Of course you can wait for saving by calling `.save()` manually, as proposed in #74 but it seems unreliable and inconcistent, if session can be updated during any request you have to use `.save()` anywhere. 
My solution may be not very nice, haven't found better way to fix this.

### DIFF
--- a/index.js
+++ b/index.js
@@ -308,10 +308,11 @@ function session(options) {
           }
 
           debug('destroyed');
+          writetop();
           writeend();
         });
 
-        return writetop();
+        return;
       }
 
       // no session to save
@@ -332,10 +333,11 @@ function session(options) {
             defer(next, err);
           }
 
+          writetop();
           writeend();
         });
 
-        return writetop();
+        return
       } else if (storeImplementsTouch && shouldTouch(req)) {
         // store implements touch method
         debug('touching');
@@ -345,10 +347,11 @@ function session(options) {
           }
 
           debug('touched');
+          writetop();
           writeend();
         });
 
-        return writetop();
+        return
       }
 
       return _end.call(res, chunk, encoding);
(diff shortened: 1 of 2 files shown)


dougwilson:
No, thanks. You're welcome yo make a fork of this module, but this PR breaks how Node.js is expected to work. You can dig through the open issues to see the reasoning for why it works the way it does today

dougwilson:
The main breakage I've seen here is that writeHead is expected to sync write out the header. This leaves no ability to queue these bytes and wait for session write before sending out the headers and thus causing a browser redirect.

Another aspect is that this also introduces Infinite buffering while the session is being saved. For example, if saving a session takes 2 seconds but in that time 500mb is written to the internal buffer here, it would be really simple to trigger a DoS.

If the solution is to buffer while the session saves, there needs to be a cap for the max amount that will be buffered before either throwing an error an canceling the response or just flushing out before the session has finished saving.

Another solution for the references issue is session locking. This is where the store can implement the ability for only a single checkout for a given session. That would mean that your issue of clicking the next button before the session saved is gone, as that second page will wait for the first to save before loading. This is common in things like ASP.net and other technologies. Unfortunately this has to be implemented in the store module, as this module has no way to track session usage that can be distributed across node instances.

andykog:
@dougwilson, thanks for your response! I just can't get your point about locking session. If you have to store some flag showing that session is locked, you must use syncronous storage (e.g. memory)? Otherwise you have no guarantee that session is locked in time before rendering second page.

Doesn't seem that [stores listed in readme](https://github.com/expressjs/session#compatible-session-stores) implement such technique, although related issues [have arised](https://github.com/voxpelli/node-connect-pg-simple/issues/31)

PR #510, opened 2017-09-23
const and arrow func

TyrfingMjolnir:
const and arrow func

The example at the bottom of the readme actually works.

### DIFF
--- a/README.md
+++ b/README.md
@@ -705,39 +705,42 @@ based session store. Supports all backends supported by Fortune (MongoDB, Redis,
 A simple example using `express-session` to store page views for a user.
 
 ```js
-var express = require('express')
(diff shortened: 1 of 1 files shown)


TyrfingMjolnir:
Not sure why this check does not go through.

gabeio:
```
/home/travis/build/expressjs/session/README.md
  715:1  error  Parsing error: The keyword 'const' is reserved
✖ 1 problem (1 error, 0 warnings)
```

dougwilson:
The checks are not working because we validate the JavaScript in the README to be syntactically correct for all Node.js versions supported by this module, at least. Arrow functions do not work in Node.js 0.10 as an example.

PR #539, opened 2018-01-10
fix getcookie cookie fallbacks for signed vs unsigned

d3zza:
It seems to me that the fallbacks for retrieving cookies for signed and unsigned cookies are implemented the wrong way around. I'm guessing that this code doesn't get executed much since these are fallbacks only triggered when cookies are not available in headers. The reason the fallbacks are necessary in my project is because I'm using express session with sockets via https://github.com/oskosk/express-socket.io-session which stores the cookies in the req.

### DIFF
--- a/index.js
+++ b/index.js
@@ -536,16 +536,7 @@ function getcookie(req, name, secrets) {
 
   // back-compat read from cookieParser() signedCookies data
   if (!val && req.signedCookies) {
-    val = req.signedCookies[name];
-
-    if (val) {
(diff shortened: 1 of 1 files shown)


dougwilson:
Hi @d3zza thanks for the pull request! I'm not sure I understand the underlying change and this breaks multiple existing tests. Can you help me understand what exactly the change is doing and is breaking the existing tests intentional?

dougwilson:
Closing per request in #540

PR #544, opened 2018-01-24
Added type check before merging options

thevoidf:


session/cookie.js
```diff
@@ -26,7 +26,11 @@ var Cookie = module.exports = function Cookie(options) {
   this.path = '/';
   this.maxAge = null;
   this.httpOnly = true;
-  if (options) merge(this, options);
+  if (options && typeof options == 'object') {
+    merge(this, options);
+  } else {
+    throw new Error('options must be a object');
```
dougwilson: This should probably be a TypeError object

thevoidf: yeah

dougwilson:
So it looks like the current test suite is not passing with these changes. Looking into it, it looks like this change is not backwards-compatible. We can certainly hold on to this for the next major version bump though 👍

PR #558, opened 2018-02-23
Update index.js

indatawetrust:
A solution to filter session usage. If req.isSessionIgnore is true, session will not be enabled. It could be a ridiculous solution, but it solved my problem. Do you think it is necessary? if you accept it, I can fill it in test and document fields. Thank you.

### DIFF
--- a/index.js
+++ b/index.js
@@ -110,6 +110,8 @@ function session(options) {
   // get the save uninitialized session option
   var saveUninitializedSession = opts.saveUninitialized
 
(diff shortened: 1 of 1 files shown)


dougwilson:
Any reason that logic can't simply be in your app directly? Or a generic middleware on npm to allow conditional exclusion of any middleware based on some condition? It seems generic enough to provide a way to generalize this to any middkeware instead of needing to PR this tyoe of change to every middleware you need it for.

indatawetrust:
You're right. Thank you👍

Issue #432, opened 2017-02-21
TypeError: option expires is invalid


spencer-rig:
#Hi,
I noticed a slight problem in trying to get my cookies to have the proper expiry.

Here's the config I'm starting with:
```js
server.use(session({
  store: new pgSession({
    pg: pg,
    conString: Config.connectionString,
    tableName: 'session'
  }),
  secret: Config.session.secret,
  resave: false,
  saveUninitialized: false,
  cookie: { httpOnly: true, secure: Config.session.secure}
}));
```

- When I set `cookie.maxAge` to an integer, the cookie ends up with that integer in `cookie.originalMaxAge` **and** in `cookie.expires` (this triggers the `option is invalid` error--`cookie.expires` should contain a Date).
- When I set `cookie.expires` to a Date, the cookie never even reaches the store.

Am I missing something? Some other way to set a cookie's max age or expiration date?

dougwilson:
Hi @spencer-rig that's weird, because almost every single test in our test suite sets the `cookie.maxAge` option, so I don't think any tests would pass if it caused an error. Perhaps there is something going wrong in your setup? Can you provide all the following?

1. The full stack trace for the error message you put in the title.
2. The version of this module you're using
3. The version of Node.js you're using

Thanks!

spencer-rig:
I just now realize that perhaps I'm setting `maxAge` to a string...

My node version is v6.9.0. Dependencies in my package.json:

```
"bluebird": "^3.4.7",
"connect-pg-simple": "^3.1.2",
"dotenv": "^4.0.0",
"express": "^4.14.1",
"express-flash": "0.0.2",
"express-session": "^1.15.0",
"github": "^8.1.1",
"lodash": "^4.17.4",
"passport": "^0.3.2",
"passport-github2": "^0.1.10",
"pg": "^6.1.2",
"pg-query": "^0.11.0",
"pug": "^2.0.0-beta11"
```
The record in my pg store:
```
OuLpY4VWT4vCF4Yto41dRHFQdIPK1LNo | {"cookie":{"originalMaxAge":"1209600000","expires":"1209600000","secure":false,"httpOnly":true,"path":"/"},"passport":{"user":{"username":"spencer-rig"}}} | 2017-02-22 15:23:13
```

And the stack trace: 
```
TypeError: option expires is invalid
    at Object.serialize (/Users/spencer/Sites/devops/mamoswine/node_modules/cookie/index.js:145:13)
[...]

PR #574, opened 2018-03-31
Added fix for maxAge string unhandled type error

PaulRBerg:
This solves #432. I'm setting the `maxAge` with an environment variable which is parsed as a string. The PR adds a small fix to automatically convert strings to integers, as the error message is vague and creates confusion.

### DIFF
--- a/test/cookie.js
+++ b/test/cookie.js
@@ -93,6 +93,12 @@ describe('new Cookie()', function () {
         assert.ok(maxAge.getTime() - Date.now() - 1000 <= cookie.maxAge)
         assert.ok(maxAge.getTime() - Date.now() + 1000 >= cookie.maxAge)
       })
+
+      it('should reject invalid types', function() {
+        assert.throws(function() { new Cookie({ maxAge: '42' }) }, /maxAge/)
+        assert.throws(function() { new Cookie({ maxAge: true }) }, /maxAge/)
+        assert.throws(function() { new Cookie({ maxAge: function () {} }) }, /maxAge/)
+      })
     })
 
     describe('path', function () {
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,6 +1,7 @@
 unreleased
 ==========
 
+  * Catch invalid `cookie.maxAge` value earlier
   * Fix issue where `resave: false` may not save altered sessions
   * Use `safe-buffer` for improved Buffer API
   * Use `Set-Cookie` as cookie header name for compatibility
--- a/session/cookie.js
+++ b/session/cookie.js
@@ -77,6 +77,10 @@ Cookie.prototype = {
    */
 
   set maxAge(ms) {
+    if (ms && typeof ms !== 'number' && !(ms instanceof Date)) {
+      throw new TypeError('maxAge must be a number or Date')
+    }
+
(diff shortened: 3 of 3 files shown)


PaulRBerg:
Just implemented the changes now! I thank you for this amazing module.

dougwilson:
Awesome, thank you! The indention seems all messed up in the changes, and I'm not sure why the CI didn't catch that. I need to take a look into this as soon as I get to a computer 👍

PaulRBerg:
I submitted again now, with proper tabs. Thing is that while I was coding, eslint shouted that it wants spaces. Perhaps just some cross-machine issue.

dougwilson:
Yea, indentation is supposed to be spaces.

test/cookie.js
```diff
@@ -39,6 +39,17 @@ describe('new Cookie()', function () {
       assert.throws(function () { new Cookie(function () {}) }, /argument options/)
     })
 
+    it('should accept integer maxAge options', function () {
+      var cookie = new Cookie({ maxAge: 7200000 })
+		  assert.equal(typeof cookie.maxAge, 'number')
+		  assert.equal(cookie.maxAge % 1, 0)
+	  })
```
dougwilson: There is definitely something wrong with either eslint for not catching this or Github is not displaying this right. This closing brace should be aligned with the "it"

test/cookie.js
```diff
@@ -39,6 +39,17 @@ describe('new Cookie()', function () {
...
+
+    it('should reject non-integer maxAge options', function() {
+	    assert.throws(function() { new Cookie({ maxAge: '7200000'}) })
```
dougwilson: And these inner lines should only br 2 spaces indented from the prior level. Is it right on your side and it is just that github is messed up? Because eslint should be catching this issue if it's really there which is very confusing.

dougwilson:
Ok, I figured out what the indentation issue is. I updated `master` so `eslint` will now detect the issue 👍

PaulRBerg:
I updated `.eslint` and repushed now. However, the issues from above simply do not appear on my machine, even after I added the rule. It might be just GitHub.

dougwilson:
Now github looks right. Did you make any changes besides adding the eslint rule?

PaulRBerg:
I did, but the code didn't look as above.

dougwilson:
> but the code didn't look as above

Hm?

PaulRBerg:
I mean it was totally fine, except it had mixed tabs with spaces. Perhaps GitHub handles this case in a special way.

dougwilson:
When I was on my machine, it looked just as messed up in my editor as it did in Github, which is how I found out what was happening. So it wasn't just github.

PaulRBerg:
Haha then I'll never use Atom again! I'm usually working in IntelliJ, but I wanted to separate this project now.

PaulRBerg:
Any updates on this? Checked out the project in IntelliJ and the indentation looks good.

dougwilson:
I am rebasing the PR now. Apparently `maxAge` can actually be a `Date` object. I'll update the PR to let that through and the next minor I'll just deprecate that feature to be removed in the future, since it's a weird feature.

PR #582, opened 2018-05-08
Consider whether request is already finished

lezsakdomi:
Fixes https://github.com/amekkawi/express-ws-routes/issues/7

### DIFF
--- a/index.js
+++ b/index.js
@@ -414,6 +414,13 @@ function session(options) {
 
     // determine if session should be saved to store
     function shouldSave(req) {
+      // Sometimes requests are already finished
+      // (see https://github.com/amekkawi/express-ws-routes/issues/7)
+      if (req.finished) {
+        debug('session ignored because request is already finished');
+        return false;
+      }
+
       // cannot set cookie without a session ID
       if (typeof req.sessionID !== 'string') {
         debug('session ignored because of bogus req.sessionID %o', req.sessionID);


gabeio:
I don't think this is a good idea. Mainly because there are reasons why you might want to adjust the session after the request is over, sessions _shouldn't_ need to update the cookie in order to update the  session since the session is mainly in the database and the cookie is only a random number key to point to a row/doc in the database.

gabeio:
@dougwilson thoughts?

lezsakdomi:
@gabeio I think my changes are now more relavant. Please review again!

dougwilson:
Just from the issues within this module I can say this will be breaking. I've seen people show their uses-cases where they reply to a request but still keep updating the session for whatever reason (file processing, etc.).

I'm saying there is nothing to change, but even looking at the original issue there really isn't any detail (plus no tests added here). Perhaps someone can provide code that demonstrates the issue itself and we can help figure out what the root cause of it is and determine what needs to happen.

Every call to `res.end()` needs to have a call to `writeend()` internally, and this cuts that off for whatever reason. We can help figure out the solution if you can help us understand the original issue (ideally with code to run that reproduces it so we can debug through what is happening).

I'm 👎 to the change as-is due to the two reasons above.

lezsakdomi:
Aaand it didn't even put an end to the problem. Maybe will provide a better POC code later.
So please note that the second commit "doesn't work".

Issue #614, opened 2018-10-04
improve session hashing when detecting modified sessions


skarbovskiy:
There is an issue in hash function
https://github.com/expressjs/session/blob/master/index.js#L582
simple `JSON.stringify` is not safe to use as when session contains an object with multiple properties inside (`user` for example with `id`, `name`, `position` properties). JSON.stringify can return different string (`id, name, position` or `position, id, name`). We need to do safe object keys sorting before serializing or use some kind of safe object hashing (e.g. https://github.com/puleos/object-hash)

skarbovskiy:
simple concept https://github.com/expressjs/session/pull/615

dougwilson:
Yea, currently the implementation does give false positives. This idea here is that it's better to just save again on accident than to accidentally not save when there was a change (i.e. false positives are ok vs false negatives are really bad). Ideally it would have no false anything, though. Efforts along those lines are welcome!

HarshithaKP:
@skarbovskiy @dougwilson  Is the PR merged into this repo or are there any pending actions ?

skarbovskiy:
in some synthetic tests object-hash implementation gives false positive "objects are identical"
https://github.com/expressjs/session/pull/615#issuecomment-440334937
so i ended up just forking express-session and implementing this fix there.
ideally we need another object-hash implementation that wont give false positives in any cases

gireeshpunathil:
were there reports of the existing hash function producing wrong results and causing issues in production deployments? I am yet to find one, is this one such a report? If not I would suggest we close this (and the associated PR), and revisit this later when we face issues.

skarbovskiy:
no, only for cases when session data use the same structure as object-hash internally (that is very unlikely)

davidmashe:
No activity for some time, and issue is debatable as an edge case. Will close for now, happy to re-open if needed.

PR #615, opened 2018-10-04
Proof of concept for #614 resolving

skarbovskiy:


package.json
```diff
@@ -14,6 +14,8 @@
     "cookie-signature": "1.0.6",
     "debug": "2.6.9",
     "depd": "~1.1.2",
+    "lodash.omit": "^4.5.0",
+    "object-hash": "^1.3.0",
```
dougwilson: Pin these to exact version plz

skarbovskiy:
@dougwilson done

test/session.js
```diff
@@ -897,9 +897,15 @@ describe('session()', function(){
   })
 
   describe('rolling option', function(){
-    it('should default to false', function(done){
+    it('should default to false and not roll cookie', function(done){
```
dougwilson: This should be it's own test, rather than altering an existing test. I can always help if you're not sure how to make one and that's why an existing test was altered vs a new one.

skarbovskiy:
@dougwilson

skarbovskiy:
anybody?)

dougwilson:
Sorry, I will take a look today.

dougwilson:
So I threw this PR into a couple of apps this morning and one of them threw an error. I reduced it down to the following test case:
```js
var express = require('express')
var expressSession = require('./index')

var app = express()

app.use(expressSession({
  saveUninitialized: false,
  secret: 'keyboard cat',
  resave: false
}))

app.get('/', function (req, res) {
  req.session.visitor = new String('username')
  res.end()
})

app.listen(3000)
```
```
$ node app.js &
$ curl -i http://127.0.0.1:3000/
TypeError: Data must be a string or a buffer
    at Hash.update (crypto.js:99:16)
    at write (project/node_modules/express-session/node_modules/object-hash/index.js:167:22)
    at Object._string (project/node_modules/express-session/node_modules/object-hash/index.js:301:7)
    at Object._object (project/node_modules/express-session/node_modules/object-hash/index.js:214:30)
    at Object.dispatch (project/node_modules/express-session/node_modules/object-hash/index.js:185:30)
    at project/node_modules/express-session/node_modules/object-hash/index.js:246:18
    at Array.forEach (<anonymous>)
    at Object._object (project/node_modules/express-session/node_modules/object-hash/index.js:242:21)
    at Object.dispatch (project/node_modules/express-session/node_modules/object-hash/index.js:185:30)
    at hash (project/node_modules/express-session/node_modules/object-hash/index.js:128:10)
```

This same app works in the current versions of `express-session`, so this would be breaking. I took a look and it seems to really just be a bug in the `object-hash` module and doesn't seem to be intended to throw like this. We could probably get a fix into `object-hash`. I can do this if you like, or would you like to to this?

But this now has me a bit worried. I wonder what other types will have this issue?

skarbovskiy:
@dougwilson, finally I got my PR fixing this issue merged in `object-hash` repo.

skarbovskiy:
@dougwilson ?)

Issue #425, opened 2017-02-09
Request: Option for refreshing the session ID


gk0us:
Sometimes, there is the need to refresh the session ID **without** loosing the session data.

Examples: 
1. Refreshing session ID after authentication (to protect against session fixation attacks)
https://www.owasp.org/index.php/Session_fixation
https://github.com/jaredhanson/passport/issues/192
2. Manually refreshing session ID before it expires (e.g. if the user wants to keep working after the maximum session lifetime, but we do not want the same session ID to be used)

dougwilson:
I don't see any particular reason something like this cannot be added here. Perhaps as an option to the current `req.session.regenerate` (like `{copy: true}`?).

gk0us:
I cannot see any reason why not, either. The `{copy:true}` option is an efficient solution.

joh-klein:
Upon a first glance, isn't this addressed with PR #419?

joh-klein:
I quite like how the did it in PHP (I know, the horror): http://php.net/manual/en/function.session-regenerate-id.php

default is to keep the old session data

joh-klein:
How would you go about it? Just copy the whole `req.session` to temp and then back again?

gk0us:
@joh-klein, I am not sure if this will work, since the old `session.id` will be copied back too. Correct me if I am wrong.

I've found this stack exchange solution but I was not able to figure it out.
http://stackoverflow.com/a/30468384

joh-klein:
If I understand [this](https://github.com/expressjs/session/blob/master/session/session.js#L26) correctly, `id` is not actually part of the session object, which correlates with my experience when I `console.log(req.session)`

dougwilson:
Hi @joh-klein `req.session.id` is indeed a thing, it's just the `id` property is not enumerable, so `console.log` just doesn't show it by default.

joh-klein:
@dougwilson , thanks for the clarification. So far, this has worked for me: 

```
let tempSession = req.session;

req.session.regenerate((err) => {
  req.session = Object.assign(req.session, tempSession);
  res.redirect('/');
});
```

I have thought about deleting `tempSession.cookie` since in my case I am also fine with a session restart.

gk0us:
Great idea @joh-klein using the `Object.assign()` function. I wasn't familiar with it. Since it only copies the enumerable properties of the `session` object, it does not mess with the refreshed `session.id` . 

[...]

PR #627, opened 2018-12-18
Use a nonce for cookie regeneration

mkawalec:
This implements https://github.com/expressjs/session/issues/425 by setting up an extra nonce cookie when `regenerate` is set to `true`. That nonce is validated against the previous value and refreshed on each request. This adds an additional write per request to the cookie store, regardless of if the state was modified or not. 

Also fixes the `getCookie` method in tests to look at all the cookies set in a response, not just the first one.

### DIFF
--- a/test/session.js
+++ b/test/session.js
@@ -896,6 +896,72 @@ describe('session()', function(){
     })
   })
 
+  describe('regenerate option', function () {
+    it('should set the session nonce', function (done) {
+      request(createServer({ regenerate: true }))
+      .get('/')
+      .expect(shouldSetCookie('connect.sidnonce'))
+      .expect(shouldSetCookie('connect.sid'))
+      .expect(200, done)
+    })
+
+    it('sets new nonce on a second request', function () {
+      const server = createServer({ regenerate: true });
+
+      return request(server)
+      .get('/')
+      .expect(200)
+      .then(function (res) {
+        const firstNonce = getCookie(res, 'connect.sidnonce');
+        return request(server)
+          .get('/')
+          .expect(200)
+          .then(function (res) {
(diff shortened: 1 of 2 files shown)


test/session.js
```diff
@@ -15,7 +15,7 @@ var Cookie = require('../session/cookie')
 var min = 60 * 1000;
 
 describe('session()', function(){
-  it('should export constructors', function(){
+  it('should export varructors', function(){
```
brian-learningpool: Find and replace error here possibly?  The test should be left as: "should export constructors".

test/session.js
```diff
@@ -896,6 +896,81 @@ describe('session()', function(){
...
+      .get('/')
+      .expect(shouldSetCookie('connect.sidnonce'))
+      .expect(shouldSetCookie('connect.sid'))
+      .expect(200, done)
+    })
+
+    it('sets new nonce on a second request', function (done) {
+      var server = createServer({ regenerate: true });
+
+      console.log('blah');
```
brian-learningpool: Did you mean to leave this in?

test/session.js
mkawalec: haha, best s// error ever

mkawalec:
@dougwilson @brian-learningpool rereview please

mihir83in:
Hi Guys any news on this ?

brian-learningpool:
@mkawalec, apologies for the delay in this, I completely missed the notification that you had updated the PR.  Your changes look good but there are conflicts in a couple of files.  

Also, I'm just a long-time fan and user of express-session, I don't have contributor access so I'm not sure what my approval is worth.

mkawalec:
Will rebase soon, hopefully we'll get this across the finish line in 2019 :D

PR #630, opened 2019-01-19
Create and apply eslint rule for semicolon enforcement

kanwisher:
I found inconsistent code styling across most of the javascript files in regards to semi-colons. Some statements ending in semi-colons, others were not. For example:

`var cookie = require('cookie');`
 `var crypto = require('crypto')` 

  This is a pull request to create consistency across the project. This change includes an eslint rule that applies a warning if a semicolon is not used at the end of a statement.

I suppose a conversation could be had where it is determined that **no** semi-colons are used. In any case, a rule should be applied and consistency should be maintained.

### DIFF
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -5,3 +5,4 @@ rules:
   no-mixed-spaces-and-tabs: error
   no-trailing-spaces: error
   one-var: ["error", { "initialized": "never" }]
+  semi: 1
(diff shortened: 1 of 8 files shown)


dougwilson:
Thanks! But this is likely a duplicate of #324

dougwilson:
The issue with landing a giant churn like this is that it will immediately make every open PR unmergable. There is inconsistent style because new coded is added in StandardJS style while PRs are getting landed.

PR #634, opened 2019-02-21
Fix bug in cookie store

geekjob-ru:
I’m using SOA (Service Oriented Architecture) application. There I’m having different services that wroted on different languages. For store session I’m using Redis. In PHP and others applications I'm implementing express-session like handlers for share sessions between the services.

But I’m also using third party services, and they can generate broken session record without cookie section. If this section is missed, then all my Node.js applications are went down with error:

```js
node_modules/express-session/session/store.js:87
var expires = sess.cookie.expires
                         ^
TypeError: Cannot read property 'expires' of undefined at RedisStore.Store.createSession
```

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -2,6 +2,7 @@ unreleased
 ==========
 
   * Fix error passing `data` option to `Cookie` constructor
+  * Fix uncaught error from bad session data
 
 1.16.0 / 2019-04-10
 ===================
--- a/index.js
+++ b/index.js
@@ -363,6 +363,19 @@ function session(options) {
       wrapmethods(req.session);
     }
 
+    // inflate the session
+    function inflate (req, sess) {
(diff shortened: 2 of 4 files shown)


dougwilson:
Hi @geekjob-ru sorry, I thought I replied to this (not sure why I thought that). Thank you for this pull request! Just a couple things to comment about:

1) We just need to add at lest one test that encounters this condition so it can be tested for to prevent regression. If you're not sure how to do this, please let me know and I can help you make a test 👍 

2) There likely shouldn't be a hard-coded magic constant that users cannot change in the code `36e5`.

3) And finally, this brings up the question on if this is actually what should happen in this condition or not; as in should what you did here be the right course of action or should this module treat it as an error (a catchable error, though) or should this module treat this condition as if the session is not present in the store?

Issue #607, opened 2018-08-28
Feature request, return promises


Soviut:
In conjunction with the callbacks, it would be nice if `session.save()` returned a promise as well. This would make in-lining session operations with if blocks, try/catch blocks and loops a lot more straightforward since async/await could be used.

For example, I'm running into race conditions with redirecting happening too soon. So in situations where I need to set the session, like a "flash" message, I have to do this:

    try {
      // some error prone operation
    } catch(err) {
      flash('error', 'something went wrong')
    }
    req.session.save((err) => {
      res.redirect('/things')
    })

You'll notice `session.save()` is called regardless of whether the error occurs. I'm choosing to do that instead of duplicate the `res.redirect()` code into multiple places.

However, what I'd rather have is:   

    try {
      // some error prone operation
    } catch(err) {
      flash('error', 'something went wrong')
      await req.session.save()
    }
    res.redirect('/things')

Slasher4k:
in the meantime could you do something like:

try {
  // some error prone operation
} catch(err) {
  flash('error', 'something went wrong')
  await new Promise((resolve,reject) =>{
    req.session.save((err)=>{
    resolve();
    })
  })
}
res.redirect('/things')

brettz9:
It'd be nice to have promises returned elsewhere where callbacks are used as well, e.g., `destroy`.

sangw0804:
i wish promise applied!! :)

Soviut:
@RussCoder No matter what, the operation is async so a return value of anything other than a promise seems kind of pointless. I've seen it done where returning a promise AND calling the callback are possible if supplied.

PR #635, opened 2019-02-24
Add promise support for sessionStore functions

brandonmanke:
Potential solution to #607.

I can add some more tests to improve coverage if necessary.

### DIFF
--- a/index.js
+++ b/index.js
@@ -370,16 +370,31 @@ function session(options) {
 
       function reload(callback) {
         debug('reloading %s', this.id)
-        _reload.call(this, function () {
-          wrapmethods(req.session)
-          callback.apply(this, arguments)
-        })
+        if (callback) {
+          _reload.call(this, function () {
+            wrapmethods(req.session)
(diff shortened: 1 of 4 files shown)


brandonmanke:
@dougwilson I can add the `should error without callback` test for `.reload()` if that is ok? Also how would you recommend explicitly testing for errors with and without callbacks? Since the tests themselves will be almost identical.

dougwilson:
Hey, so I added a couple more conditions on Monday to further bump the coverage to help out 👍Feel free to add some new ones as well; I just pushed up without the rebase on Monday just in case you were working on some as well.

I took a look at the coveralls config and it's set only to care about line coverage. So I wouldn't really worry about the `reject` branches to get it landed. If you want to add them anyway as an exercise, I would just copy-paste the tests at this time.

brandonmanke:
How would you recommend testing https://coveralls.io/builds/21919265/source?filename=index.js#L385?

bjohansebas:
closes in favor of #737

PR #639, opened 2019-02-28
Fixed res.send() bug.

ZenMittal:
res.send() was mistakenly written as res.end() in two places in an example code.


dougwilson:
Hi! And thanks for the pull request. In this case res.end is the correct method. When using express and a call is made to res.write, you cannot then call res.send , only res.end to end the response.

The difference is that res.send you give a string that is the entire body, while res.write and res.end lets you call multiple res.writes with the body and it is finalized with a call to res.end

PR #643, opened 2019-03-11
Use 'req.sessionStore' instead of directly using 'store' object.

sumeetkakkar:
## Description
Use `req.sessionStore` instead of directly using `store` object. This gives flexibility to inject additional parameters like `req` to `store.get` and `store.destroy` via custom traps using [JavaScript Proxy](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy) without changing the interface of the `Store`.

## Motivation and Context
We want to pass `req` object down to the methods of our `Store` implementation. `store.set` method can access it from `session.req` but it is not available in `store.get` and `store.destroy` methods.

One workaround we found is to leverage setter of `req.sessionStore` to create [JavaScript Proxy](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy) of the `store` object. This proxy has traps for `store.get` and `store.set` and it adds the `req` object as a parameter to the function call.

To make this work we need changes in this module to use `req.sessionStore.get` and `req.sessionStore.destroy` methods so that the proxy object is used.

### DIFF
--- a/index.js
+++ b/index.js
@@ -303,7 +303,7 @@ function session(options) {
       if (shouldDestroy(req)) {
         // destroy session
         debug('destroying');
-        store.destroy(req.sessionID, function ondestroy(err) {
+        req.sessionStore.destroy(req.sessionID, function ondestroy(err) {
           if (err) {
             defer(next, err);
           }
@@ -458,7 +458,7 @@ function session(options) {
 
     // generate the session object
     debug('fetching %s', req.sessionID);
-    store.get(req.sessionID, function(err, sess){
+    req.sessionStore.get(req.sessionID, function(err, sess){
       // error handling
       if (err) {
         debug('error %j', err);


knoxcard:
Hmmm, I like where you are going with this and I have a solution to this as well.

knoxcard:
What I have been pondering...

```
  var express_session = require('express-session'),
    redis_store = new (require('connect-redis')(express_session))()
  app.set('redis_store', redis_store)
  app.session = express_session({
    store: redis_store,
    secret: process.env.session_secret,
    name: process.env.session_name,
    rolling: true,
    saveUninitialized: true,
    unset: 'destroy',
    resave: true,
    proxy: true,
    logErrors: false,
    cookie: {
      path: '/',
      domain: '.' + process.env.app_domain,
      httpOnly: true,
      secure: process.env.protocol === 'https',
      maxAge: (60 * 60 * 1000) // 60 mins
    }
  })
```

sumeetkakkar:
Thanks for looking into this. 
The idea behind this PR is to be able to trace the logging for a `req`. Especially to know why error happened while processing the request and one source of the error is the sessionStore. Right now we are using https://www.npmjs.com/package/continuation-local-storage and we want to move away from it.

A `Store` implementation which can enable us to do this is:
```
const Store = app.Store || app.session.Store;
class SomeStore extends Store {
  get(id, req, callback) {
    if (typeof req === 'function') {
      callback = req;
      req = void 0;
    }
    ....
    if (req) req.log(...);
    ....
  }
  set(id, session, callback) {
    ....
    if (session.req) session.req.log(...);
    ....
  }
  destroy(id, req, callback) {
    if (typeof req === 'function') {
      callback = req;
      req = void 0;
    }
    ....
    if (req) req.log(...);
    ....
  }
  ...
}
```

Please note that this is a modified interface and changing `store.get(req.sessionID, function(err, sess){` (https://github.com/expressjs/session/blob/master/index.js#L474) will pretty much break everybody. The workaround we are exploring is to inject `sessionStore` getter and setter in `req` object and leverage `req.sessionStore = store;` (https://github.com/expressjs/session/blob/master/index.js#L214) in the `express-session` middleware function. 

```
Object.defineProperty(req, 'sessionStore', { 
    get: function() { return this[kSessionStore]; },
    set: function(store) { 
        return this[kSessionStore]=constructJSProxy(store, this);
    } 
});
```
`req.session.save` and `req.session.destroy` are already using `this.req.sessionStore`.

The JS Proxy looks like following:
```
   const constructJSProxy = function constructSessionStoreProxy(store, req) {
        const handler = {
            get: function(target, prop, receiver) {
                const origMethod = target[prop];
                if (typeof origMethod !== 'function') {
                    return origMethod;
                }
                switch (origMethod) {
                    case SomeStore.prototype.get:
                    case SomeStore.prototype.destroy:
                        return function (id, callback) {
                            return origMethod.call(target, id, req, callback);
                        };
                    default:
                        return origMethod;
                }
            }
        };
        return new Proxy(store, handler);
    };
```

sumeetkakkar:
@knoxcard, Please let me known if there are any specific concerns you want me to address to help merge this PR.

As I have stated, our goal is to be able to access `req` object in the Store's `get` and `destroy` methods. Any alternate recommendation to achieve this is also appreciated.

knoxcard:
@sumeetkakkar - I am not a project contributor or owner to this repo.  Your best bet is having @dougwilson review this.

dougwilson:
Can you look at #712 and confirm if that will accomplish the same thing as here? It is further along and seems much more flexible, including keeping our internals internal and less at risk for user overrides causing hard to debug issues.

sumeetkakkar:
> Can you look at #712 and confirm if that will accomplish the same thing as here? It is further along and seems much more flexible, including keeping our internals internal and less at risk for user overrides causing hard to debug issues.

👍

sumeetkakkar:
@dougwilson , Can you please share ETA for #712?

dougwilson:
We are working though everything, just there is a large backlog. Part of resolving that is resolving the difference between these two PRs, which is why I messaged on here.

Were you able to take a look at the changes in #712 ? Do you believe they will accomplish the same thing has this pull request?

sumeetkakkar:
Our goal is to be able access `req` object in store.get and store.destroy. It seems #712 will make it possible. So yes this PR can be closed in favor of #712.

Thank you!

PR #649, opened 2019-04-11
TypeError: Cannot set property data of #<Object> which has only a getter fix

eoL95:
Hello. In new release you have an error:
`TypeError: Cannot set property data of #<Object> which has only a getter`

So, this happened because you've changed this code 
image
But in your `cookie.js` file `strict mode` is enabled. That is the reason why JS throws an error.
So I think the better way is to define setter for data field in prototype.

### DIFF
--- a/session/cookie.js
+++ b/session/cookie.js
@@ -33,7 +33,9 @@ var Cookie = module.exports = function Cookie(options) {
     }
 
     for (var key in options) {
-      this[key] = options[key]
+      if (key !== 'data') {
+        this[key] = options[key]
+      }
     }
   }
 
--- a/test/cookie.js
+++ b/test/cookie.js
@@ -44,6 +44,15 @@ describe('new Cookie()', function () {
       assert.throws(function () { new Cookie(function () {}) }, /argument options/)
     })
 
+    it('should ignore "data" option', function () {
+      var cookie = new Cookie({ data: { foo: 'bar' }, path: '/foo' })
+
+      assert.strictEqual(typeof cookie, 'object')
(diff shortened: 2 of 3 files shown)


test/cookie.js
```diff
@@ -37,6 +37,14 @@ describe('new Cookie()', function () {
       assert.strictEqual(typeof new Cookie({}), 'object')
     })
 
+    it('should create cookie object with data field in options', function () {
+      assert.doesNotThrow(function () {
+        var cookie = new Cookie({
+          data: {}
```
dougwilson: Why would your code be passing in an option that is simply discarded? If this is what is causing the error, ideally you should remove the part from your code that is passing in this option.

session/cookie.js
```diff
@@ -124,6 +124,8 @@ Cookie.prototype = {
     }
   },
 
+  set data(data) {},
```
dougwilson: Based on the test provided, calling this should show a deprecation message to alert users this will be removed in the version 2.0

dougwilson: I updated the PR to remove the setter.

test/cookie.js
eoL95: Yes, it's true. But function shouldn't break if it accepts some extra data. I think your next comment is good.

dougwilson: Right, but the setting controls more than just passing in more data to the function. For example it now enables `req.session.data = 'foo'` which would have been an error even in 1.15.x.

session/cookie.js
eoL95: Seems good

PR #656, opened 2019-04-26
Import & update typings from DefinitelyTyped

HoldYourWaffle:
The current [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/express-session/index.d.ts) typescript definitions for this module are a mess. They're needlessly convoluted and in some ways outdated or wrong.
Because my definitions would be a breaking change and because DT is a mess in general (giant mono-repo, having to install another package) I suggest merging these definitions into the main repository.

I made 2 commits to make the difference between the DT definitions and mine clearer, I'd be happy to squash them before this is merged.

### Changes relative to the DT definition
- `regenerate`, `load` and `createSession` are not functions on `Store`. I'm not sure why they were there in the first place, I think `Session` has *similar* methods but they aren't a proper match.
- Reuse [`Express.CookieOptions`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/56607c79bd4f6de9a8d77925e6e09aaf0cf6491a/types/express-serve-static-core/index.d.ts#L165)  for `SessionCookie`¹
- `SessionData` no longer has a `string => any` index type. This makes it possible to use [declaration merging](https://www.typescriptlang.org/docs/handbook/declaration-merging.html) to allow the user to type their own sessions. This is also seen on express's [Request](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/56607c79bd4f6de9a8d77925e6e09aaf0cf6491a/types/express-serve-static-core/index.d.ts#L16) type.
- There is no `BaseMemoryStore`, `MemoryStore` just extends `Store`
- A less convoluted type hierarchy

**Notes**
- I suggest removing the optional modifier on `Express.Request.session` and `Express.Request.sessionID`. Although they're technically 'optional' in the sense that they don't exist before you `use` your session middleware it's really inconvenient when working with strict null checking enabled. If you have `use`d the middleware these properties are guaranteed to be there, making any checks redundant. Because of this I'd suggest going for convenience rather than accuracy in this instance.
- ¹I'm definitely not sure about the types of the cookie related properties. There seems to be a container `Cookie` class [here](https://github.com/expressjs/session/blob/327695e3d1dd4d7a5d3857f2afbb5852b7710213/session/cookie.js#L25) but I'm not sure how it's used and how to represent it in these definitions. I made it an interface which extends [`Express.CookieOptions`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/56607c79bd4f6de9a8d77925e6e09aaf0cf6491a/types/express-serve-static-core/index.d.ts#L165) for now but I'm pretty sure this neither accurate nor semantically correct.

<br><br>

@horiuchi, @jacobbogers, @builtinnya and @ry7n, you are credited as the original authors of the DT definitions. I have a couple of questions about why they are the way they are:
- Why is the type hierarchy so convoluted? I couldn't find a reference to anything resembling a `BaseMemoryStore` and several inheritances redefined common functions.
- What do you think the cookie types should be?
- Are there actually good reasons for why the convoluted hierarchies are that way, in other words: did I miss something?


HoldYourWaffle:
I just noticed that although the readme says that [a store *must* be an `EventEmitter`](https://github.com/expressjs/session#session-store-implementation) there's no mention of *why* they have to be and how it's actually used.
I've never used `EventEmitter`s before, but from what I can tell by a quick google search there doesn't seem to be a reason for this requirement. Even the [connect-redis](https://github.com/tj/connect-redis) implementation that's set as an example doesn't seem to implement it (although I could just be missing something).

Is the readme wrong or is being an `EventEmitter` an actual requirement? If so, why do so many stores not implement it (including the one used as example in the readme)?

HoldYourWaffle:
Turns out declaration merging doesn't work properly with these definitions, currently investigating.

index.d.ts
```diff
@@ -0,0 +1,75 @@
...
+
+interface Session extends SessionData {
+  id: string;
+  cookie: SessionCookie;
+
+  regenerate(callback: (err: any) => void): void;
+  destroy(callback: (err: any) => void): void;
+  reload(callback: (err: any) => void): void;
+  save(callback: (err: any) => void): void;
+  touch(callback: (err: any) => void): void;
```
dougwilson: This method does not take a callback.

index.d.ts
```diff
@@ -0,0 +1,75 @@
...
+declare function session(options?: SessionOptions): express.RequestHandler;
+export default session;
+
+
+
+interface Session extends SessionData {
+  id: string;
+  cookie: SessionCookie;
+
+  regenerate(callback: (err: any) => void): void;
```
dougwilson: All these methods seem to indicate void return, though there is a return value: the session object.

index.d.ts
```diff
@@ -0,0 +1,75 @@
...
+  destroy(callback: (err: any) => void): void;
+  reload(callback: (err: any) => void): void;
+  save(callback: (err: any) => void): void;
```
dougwilson: The callback is optional for this method, however you indicate that in typescript.

index.d.ts
```diff
@@ -0,0 +1,75 @@
...
+  touch(callback: (err: any) => void): void;
+}
+
+declare interface SessionCookie extends express.CookieOptions { //XXX actually a class
+  originalMaxAge: number;
+}
+
+interface SessionData { } //seperate interface to allow application-specific declaration merging
```
dougwilson: Likely needs documentation about this in the readme.

HoldYourWaffle:
I have addressed your concerns, including a crude concept of additional documentation in the readme.

What do you think about my suggestion to remove the optional modifier on `Request.session` and `Request.sessionID`? Seems like a reasonable trade-off to me, especially with a note on it in the readme (included in the current concept).
And what about the `EventEmitter` requirement? I haven't found a reason why it's needed nor a store that actually fulfills it.

dougwilson:
So i have not finished reviewing yet, just things I found from a first look.

> What do you think about my suggestion to remove the optional modifier on Request.session and Request.sessionID? 

I don't know what this means, can you elaborate if you need an answer from me?

> And what about the EventEmitter requirement? I haven't found a reason why it's needed nor a store that actually fulfills it.

It is required. Try making a store without it and it will fail immediately since this module attempts to add event listeners to the store during set up. Not sure what the confusion is.

index.d.ts
```diff
@@ -0,0 +1,77 @@
...
+
+
+
+interface SessionOptions {
+  secret: string | string[];
+
+  genid?(req: express.Request): string;
+  name?: string;
+  store?: Store;
+  cookie?: express.CookieOptions;
```
dougwilson: Where can I look up how this is defined? Not sure if this is a 1:1 mapping with the options in this module or not.

HoldYourWaffle: [Here](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/56607c79bd4f6de9a8d77925e6e09aaf0cf6491a/types/express-serve-static-core/index.d.ts#L165)

dougwilson: Is just the properties and their values documented in the README (https://github.com/expressjs/session#cookie) confusing? Like we can make them better, do you have any suggestions?

dougwilson: Looking at the linked type and here, I think the following are the differences:

- The `expires` is just a `Date` object, not `boolean` as defined in https://github.com/DefinitelyTyped/DefinitelyTyped/blob/56607c79bd4f6de9a8d77925e6e09aaf0cf6491a/types/express-serve-static-core/index.d.ts#L168
- As far as I'm aware there is no `encode` option accepted by this module but it is defined at https://github.com/DefinitelyTyped/DefinitelyTyped/blob/56607c79bd4f6de9a8d77925e6e09aaf0cf6491a/types/express-serve-static-core/index.d.ts#L173

index.d.ts
```diff
@@ -0,0 +1,77 @@
...
+  cookie: SessionCookie;
+
+  regenerate(callback: (err: any) => void): Session;
+  destroy(callback: (err: any) => void): Session;
+  reload(callback: (err: any) => void): Session;
+  save(callback?: (err: any) => void): Session;
+  touch(): Session;
+}
+
+declare interface SessionCookie extends express.CookieOptions { //XXX actually a class
```
dougwilson: what does the XXX comment here mean? it is a todo?

index.d.ts
```diff
@@ -0,0 +1,76 @@
...
+}
+
+interface SessionCookie extends express.CookieOptions { //XXX actually a class
+  originalMaxAge: number;
+}
+
+
+
+export abstract class Store extends node.EventEmitter {
+  constructor(config?: any);
```
dougwilson: Is this a reference to the `Store` class in this module? There are no arguments to the constructor if so: https://github.com/expressjs/session/blob/master/session/store.js#L32

dougwilson: Actually all these methods are different, so not sure what this references and then where the declaration to the `Store` class is.

index.d.ts
HoldYourWaffle: It's a note, mostly targeted at the original authors. The issue is that `SessionCookie` (which is a reference to the 'class' defined in [cookie.js](https://github.com/expressjs/session/blob/master/session/cookie.js)) is technically a class, but as far as I can tell it shouldn't be referenced directly by the user. This is mainly what I meant by not understanding the cookie system, I was hoping the original authors could shine some light on what they decided to do and why.

(@horiuchi @jacobbogers @builtinnya @Ry7n)

dougwilson: I mean, it is exported as `require('express-session').Cookie`. It's needed by stores created without the use of the `require('express-session').Store` export.

index.d.ts
HoldYourWaffle: > Is this a reference to the Store class in this module? There are no arguments to the constructor if so

It is. Seems like the DT definitions are wrong again (although I probably should've checked before copying it).

> Actually all these methods are different, so not sure what this references and then where the declaration to the Store class is.

It was meant to be an 'abstract base class' (like in Java) for Store implementations. Turns out I understood abstract classes in TypeScript wrong. I do know now where `regenerate`, `load` and `createSession` are coming from, I must've completely overlooked them at first.

I have updated the definitions with the proper abstract class definition.

index.d.ts
```diff
@@ -0,0 +1,88 @@
...
+
+  regenerate: (req: express.Request, callback: (err?: any) => any) => void;
+  load: (sid: string, callback: (err: any, session?: SessionData) => any) => void;
+  createSession: (req: express.Request, sess: SessionData) => void;
+
+  abstract get: (sid: string, callback: (err: any, session?: SessionData | null) => void) => void;
+  abstract set: (sid: string, session: SessionData, callback?: (err?: any) => void) => void;
+  abstract destroy: (sid: string, callback?: (err?: any) => void) => void;
+  abstract all: (callback: (err: any, obj?: { [sid: string]: SessionData; } | null) => void) => void;
```
dougwilson: So forgive me If I don't understand and this may be a non-issue: there are methods declared here that stores are not required to implement (https://github.com/expressjs/session#session-store-implementation). Would this cause issues if extended and they are not implemented?

index.d.ts
```diff
@@ -0,0 +1,88 @@
...
+  abstract clear: (callback?: (err?: any) => void) => void;
+  abstract touch: (sid: string, session: SessionData, callback?: (err?: any) => void) => void;
+}
+
+export class MemoryStore extends Store {
+  get: (sid: string, callback: (err: any, session?: SessionData | null) => void) => void;
+  set: (sid: string, session: SessionData, callback?: (err?: any) => void) => void;
+  destroy: (sid: string, callback?: (err?: any) => void) => void;
+  all: (callback: (err: any, obj?: { [sid: string]: SessionData; } | null) => void) => void;
+  length: (callback: (err: any, length?: number | null) => void) => void;
```
dougwilson: So the `length` will never be `null` for `MemoryStore`. Impl here: https://github.com/expressjs/session/blob/master/session/memory.js#L134 Not sure if that is what this definition is implying or not.

HoldYourWaffle:
> I don't know what this means, can you elaborate if you need an answer from me?

Typescript employs strict null/undefined checking, which basically means that the compiler screams at you for using a nullable/undefinedable reference without the proper checks. It's really inconvenient in this case, although technically correct to have the `session` and `sessionID` properties be 'undefinedable' (optional). The original DT definitions *did* list these properties as optional, but I'm pretty sure other middleware type definitions don't do this. Either way it's just really inconvenient and messy to have these properties as 'optional', though technically correct.
I hope that's enough elaboration, if not I'm sure one of the original authors will be happy to comment on this matter.

> It is required. Try making a store without it and it will fail immediately since this module attempts to add event listeners to the store during set up. Not sure what the confusion is.

Apologies, I completely missed [this](https://github.com/expressjs/session/blob/327695e3d1dd4d7a5d3857f2afbb5852b7710213/session/store.js#L40) line which effectively causes *all* store implementations to implicitly extend `EventEmitter`. I have updated the definitions accordingly.

dougwilson:
> I hope that's enough elaboration, if not I'm sure one of the original authors will be happy to comment on this matter.

Well, I've never used typescript and still this seems over my head to be able to answer.

index.d.ts
```diff
@@ -0,0 +1,88 @@
...
+  clear: (callback?: (err?: any) => void) => void;
+  touch: (sid: string, session: SessionData, callback?: (err?: any) => void) => void;
```
dougwilson: The `callback` will never return an error (impl: https://github.com/expressjs/session/blob/master/session/memory.js#L156)

index.d.ts
HoldYourWaffle: Good catch. I made `all`, `length`, `clear` and `touch` optional.

index.d.ts
HoldYourWaffle: `length?: number | null` means that there's an *optional* parameter `length` (could be `undefined`, see my earlier comment) that can be of type `number` **or** `null` (`|` defines a [union type](https://www.typescriptlang.org/docs/handbook/advanced-types.html#union-types), effectively an `or`). If you didn't understand any of that: doesn't matter, it's probably wrong in this case.

I didn't change this type from the DT one, but now that you mention it it seems a little backwards. Would a regular (non-optional) `length` parameter of type `number` make more sense?

dougwilson: So right now I'm just looking through and trying to eye ball this. I have never used typescript and don't understand it's specifics, so cannot answer anything related to it. Maybe if there were instructions for how I could use this in an example app I could play around? Not sure if that would help me answer your question or not.

index.d.ts
HoldYourWaffle: The original DT typings are even worse than I initially thought. I removed the parameter.

index.d.ts
HoldYourWaffle: I think the [Typescript in 5 minutes](https://www.typescriptlang.org/docs/handbook/typescript-in-5-minutes.html) and [Basic Types](https://www.typescriptlang.org/docs/handbook/basic-types.html) pages from the official documentation would be your best bet for setting up and understanding a basic TypeScript project. Keep in mind that you should probably set `strictNullChecks` to `true` in your `tsconfig.json` to get the best results.

To use these new definitions in your own project you could run `npm install github:HoldYourWaffle/express-session#integrate-update-typings` to install from my fork directly (although you'd have to re-run this every time I create a new commit, since npm doesn't properly support updates for git repositories yet), or you could just manually copy the [`index.d.ts`](https://github.com/HoldYourWaffle/express-session/blob/integrate-update-typings/index.d.ts) file from my fork into your own project (not sure if this will work immediately out of the box, feel free to ask for help if you can't get it to work).

Will all of this help you with answering my question? Probably not, since it's for one a simple typing choice and secondly it's a semantic choice: is it valid for implementations to pass in certain types?

To rephrase my question: is it valid for implementations to pass `null` and/or `undefined` as `length` to the callback?

HoldYourWaffle:
I found another possible mistake in the typings which boils down to the following: do the functions on the `Store` class expect a full-blown `Session` (including `id`, `cookie`, `regenerate`, etc.) or *just* the user-defined data properties?
The DT definitions went for the latter (although I could just be misinterpreting the weird convoluted hierarchy), but I'm not sure that's correct as it seems pretty backwards.

index.d.ts
HoldYourWaffle: > Is just the properties and their values documented in the README (https://github.com/expressjs/session#cookie) confusing? Like we can make them better, do you have any suggestions?

Although the formatting could be better the README is quite clear from a user perspective. It's however hard to understand how the code's "type system" actually works (mainly because it's, well, JavaScript) which is pretty important if you're writing type definitions.

> Looking at the linked type and here, I think the following are the differences: ...

Updated. Just to be sure: does `secure` indeed accept a string "auto" or did you overlook that one?

dougwilson: Yes, it accepts the string auto:

https://github.com/expressjs/session#cookiesecure

> The cookie.secure option can also be set to the special value 'auto' to have this setting automatically match the determined security of the connection.

index.d.ts
HoldYourWaffle: I have now defined the hierarchy as follows:
- There's a class `Cookie`
- This class implements `CookieOptions`, meaning it has all the properties defined by it
- The interface (basically a list of properties) `CookieOptions` is based on the README and your comments above
- `Session.cookie` has the type `Cookie`

Does this sound correct to you?

dougwilson: Yea

HoldYourWaffle:
I have decided to drop the optional modifier on `session`/`sessionID`, mainly because in 99% of the cases it's completely useless. It's just unnecessarily inconvenient, and it's safe to assume someone will actually register the middleware if they're using this module.

I also decided to give the `length` parameter of the callback passed to the function with the same name a type of just `number`. I can't think of a use-case where `length` would return anything else than a `number` without an error having occured (which in my opinion should result in an error being thrown). Although technically valid, I don't think it's worth it to bother typescript users with a `null`able/`undefined`able typing.

Lastly I took another shot at the whole cookie system hierarchy. I *think* I got it right this time, but you should definitely check my summary in the comment above in case I missed something.

Are there any more changes you'd like to see before this can get merged?

dougwilson:
Thanks so much! I will take another review though them due to the number of changes to double check as well :+1:

As for next steps, ignoring the outcome of the review, it would be nice to...

1. Have a passing CI to merge
2. Have tests in some form to, at minimum, verify that the tsd is syntacticly correct since it is hand maintained
3. Is this safe to just land in 1.x? Not sure how landing this affects existing ts users.

HoldYourWaffle:
> Have a passing CI to merge

Don't ask me how but I never noticed that the tests are failing. No idea how that can happen by the addition of typings but I'll take a look.

> Have tests in some form to, at minimum, verify that the tsd is syntacticly correct since it is hand maintained

I'll see how [DefinitelyTyped](http://definitelytyped.org/) does this, since they're the biggest (hand-maintained) repository of TypeScript definitions.

> Is this safe to just land in 1.x? Not sure how landing this affects existing ts users.

No idea actually. It's probably going to conflict somehow with the [DefinitelyTyped version](https://www.npmjs.com/package/@types/express-session) but I'm not sure how well the TypeScript compiler handles duplicate typings. It would make sense for it to only use the in-package definitions but I don't think that's actually how it works. However, if this gets merged the DefinitelyTyped package will ultimately be deprecated/removed providing existing TS users with a nice little warning that'll tell them there are new (and improved) definitions.
As for *safety*, I think these definitions are under normal usage circumstances compatible (since they both describe the same module and the DT one technically did 'work' although badly).

dougwilson:
So the safe to land in 1.x is, overall, my main concern. Is there any way to gather what would happen to an existing app?

index.d.ts
```diff
@@ -0,0 +1,110 @@
...
+  originalMaxAge: number;
+
+  maxAge?: number;
+  signed?: boolean;
+  expires?: Date;
+  httpOnly?: boolean;
+  path?: string;
+  domain?: string;
+  secure?: boolean | 'auto';
+  encode?: (val: string) => void;
```
dougwilson: I don't think there is such a property as this available.

HoldYourWaffle: I blame VS Code for auto-inserting this (╯°□°）╯︵ ┻━┻

HoldYourWaffle: Fixed

HoldYourWaffle:
> So the safe to land in 1.x is, overall, my main concern. Is there any way to gather what would happen to an existing app?

Do you know a codebase I could test on?

dougwilson:
No, I have not used TypeScript before.

HoldYourWaffle:
Turns out CI is only failing in Node v0.8, probably because it's just too old to install the `@types/express` and `@types/node` packages (although I'm not sure why those packages would cause issues).
If you don't mind I can just remove Node v0.8 from the test matrix, which I don't think will be that big of an issue since it's almost 7 years old now. [This](https://github.com/expressjs/discussions/issues/11#issuecomment-200807034) comment suggests that only 0.04% of people were using Node v0.8 **3 years ago** (not sure where this number comes from but I really don't think the actual percentage will be much higher).

dougwilson:
So our (and my) policy is that it would require a new major version to drop support for a Node.js version. We can drop it, but it would then push landing this to 2.0, placing it behind the queue of 1.x PRs to land. I'm just providing this for transparency.

HoldYourWaffle:
I did some more digging, seems like the npm version used by travis with node v0.8 is not supported anymore. Not sure why it still works for other commits, maybe travis caches installed packages?

I tested replacing DT typings with these on about 10 repositories listed on npm that depend on this package, got pretty inconclusive results. Most repo's were broken, but from what *did* work I can already tell that stuff certainly *can* go wrong. This seems to be mostly import-type related (which makes sense).

So since it seems like the major version should be bumped anyway for this, would it be alright to remove v0.8 from travis? Also, how long will it be until v2.0 is released? There's no rush, I can use my own fork in the meantime, but it's nice to have an indication. It would also help with the process of deprecating/removing the old DT definitions.

dougwilson:
> So since it seems like the major version should be bumped anyway for this, would it be alright to remove v0.8 from travis?

Yes, you are welcome to do so.

> Also, how long will it be until v2.0 is released?

I need to go through the PRs created before yours to land still. Most (if not all) can land on 1.x. I will rebase this PR into the 2.0 branch when it is cut. I don't have a specific timeline for this.

HoldYourWaffle:
That makes sense, I'll use my fork for now.

Meanwhile I'll try to convince the folks over at DT to deprecate the old definitions.

dougwilson:
And to add that since this now cannot land until 2.0, the question about landing in 1.x for existing TypeScript users is obvious irrelevant :)

package.json
```diff
@@ -20,6 +21,8 @@
     "uid-safe": "~2.1.5"
   },
   "devDependencies": {
+    "@types/express": "^4.16.1",
+    "@types/node": "^11.13.7",
```
dougwilson: A question on this: do these lock down the versions of these folks can use with TypeScript? Like this would mean Node.js 10.x users may have an issue or 12.x users, or... ? If not, what do they do?

dougwilson: I guess they are in `devDependencies` so no users would ever get them installed from this, right?

HoldYourWaffle: > A question on this: do these lock down the versions of these folks can use with TypeScript? Like this would mean Node.js 10.x users may have an issue or 12.x users, or... ? If not, what do they do?

Good question. I know that NPM does this thing where it maintains a copy of each version of a package that was required but that aren't compatible with each other, but I'm not sure how the TypeScript compiler handles these 'nested dependencies'.
As far as I know this shouldn't cause issues because:

> I guess they are in devDependencies so no users would ever get them installed from this, right?

Yup. Only direct `npm install`s from this project will download them.

dougwilson: Gotcha. So if I remove them everything is working. Is there something I'm missing or us there some kind of validation missing?

HoldYourWaffle: I'm not quite sure what you mean, but I've done some more research and updated the dependency versions accordingly. I used [peer dependencies](https://nodejs.org/es/blog/npm/peer-dependencies/) to make sure users don't get locked in specific versions for these definitions (to make development easier I listed the same dependencies as `devDependencies`, otherwise npm would only complain about missing peer-deps instead of actually installing them).
I have not locked down `@types/node` at all because I think we can be pretty confident that `EventEmitter` is not going anywhere. I locked `@types/express` to any 4.x version, is that okay?

The reason consumers of this package need these definitions is that they're exported from this module is some way (`EventEmitter` from `node` though `Store` which extends it, and `Request`, `Response` and `RequestHandler` from `express` through functions).

HoldYourWaffle: I just remembered that non-TS users will be bothered about these packages if I list them as peer dependencies so I removed them as such. TS users will still need to install these packages to properly use this package, but I think it's safe to assume everyone will have these already. Perhaps a little mention in the README would be nice?

HoldYourWaffle:
I have squashed all commits down to only 2: the actual definitions & the support drop for v0.8

package.json
dougwilson: Yea, feel free to add mention in the readme 👍

HoldYourWaffle:
In case anyone wants to use these definitions before the next major release without the hassle of git dependencies in node: I have published my fork [here](https://www.npmjs.com/package/@holdyourwaffle/express-session).

HoldYourWaffle:
It's been 8 months now, is there any chance of this getting merged soon?
Fixed typings for [downstream session stores](https://github.com/chill117/express-mysql-session/pull/93) depend on this PR, so it'd be nice to have a solution that doesn't involve [forked](https://github.com/expressjs/session/pull/656#issuecomment-489643538) [dependencies](https://github.com/chill117/express-mysql-session/pull/93#issuecomment-489652251).

Little note to myself: notify https://github.com/JLuboff/connect-mssql-v2/issues/10 when this is merged.

dougwilson:
https://github.com/expressjs/session/pull/656#issuecomment-489245845

voxpelli:
Wouldn’t it be possible to make the types optional dependencies?

They are probably failing because they are scoped dependencies that such an old npm version cannot handle?

Landing this in 1.x would be preferable as any future 2.x is just that – something that eventually happens at some point in the future.

dougwilson:
I'm not super familiar with typings, so can't really say. Besides the node.js change in this PR, I would assume if these landed in a 1.x version it would still break typescript projects, since the typings here are not compatible with existing typings folks are using, is that a correct assumption?

HoldYourWaffle:
> Wouldn’t it be possible to make the types optional dependencies?
They are probably failing because they are scoped dependencies that such an old npm version cannot handle?

I don't fully understand what you're trying to say here.

> since the typings here are not compatible with existing typings folks are using, is that a correct assumption?

I don't fully remember, but I'd be surprised if these new typings don't break *something*.

These changes definitely need to go into a new major version, but I agree with what @voxpelli says (I think). It's been 8 months since I submitted this PR, and (as far as I can tell) v2 hasn't come any closer. You say you want to wait with releasing v2 until all v1-PR's have 'landed', which makes a lot of sense, but if there's no work being done to reach this point it could take years to resolve this mess.

I'm not saying there's nothing happening, because I honestly have no idea if that's the case. I haven't used this module for several months, the only reason I checked in was because a downstream maintainer asked for a status update (https://github.com/chill117/express-mysql-session/pull/93#issuecomment-572961325).

The current typings mess is (most likely) affecting everyone who uses this library with TS, so I think it's fair to say that a quick solution would be appreciated by a lot of people. I don't want to rush you or anything, I'm just trying to help resolve this typings mess, in the hope that others won't have to go through the same troubles as me.

HoldYourWaffle:
English is not my first language, so I'm sorry if I seem a bit blunt.

Perhaps it'd be possible to release v2 before all PR's have landed?
This means that the new features won't be in v1 of course, but since the breaking changes are minimal updating shouldn't be a problem almost all, if not all, users.

dougwilson:
If you really need a quick solution, why can the typings not be updated in definatelytyped? Wouldn't that solve everything at concern here? Since that is a separate dependency folks can pull in the version of types that work for them without breaking others and then the others have a chance to migrate? Especially since the types are for the 1.x of this module.

dougwilson:
I'm ultimately not interested in releasing a major version of this module _just to fix typings I never made_. I'm happy to land them in here in conjunction with the other badly-needed changes. But I'm not sure why now the pressure is on a module and person who did not make a bunch of broken typings in the first place... this will be landed along with 2.0, but if you really need to "fix the mess" that is not even part of this module, land the fix in definitelytyped in the interim. Folks can use their package.json to refer to the relevant type definitions.

dougwilson:
Note I have locked this thread temporarily while I can get some samesite cookie releases out tonight for folks and can resume this discussion at a later time. One thing to consider for later is who will keep the types up to date with changes in this module? Landing this before the other PRs will make them unlandable as-is since the types will be off. I just want to get those out of the way before we land the types so they are at least not just immediately out of data again. This week I will start a 2.0 beta branch to start getting things landed. We will use that as a test to see if there is anyone who helps keep the types up to date though those changes in the beta as a test for them going in the final release. That is just a proposal, and we can further discuss. I will open a meta issue so things are not lost.

dougwilson:
Hi @HoldYourWaffle I have unlocked at your request, but I just want to make sure that you (not I) say anything rash is all. Many times I observe when a github issue or pr turns into a chat (based on the speed of responses) it can get out of hand (and I'm not always not at fault either). Especially because responses can be long and both people end up typing and posting at the same time, reading each other's response and again and again :) I don't want PRs just hanging around just like you expressed, but there is a lot going on in the conversation here. I hope some of the previous posts will have some clarity, and I'm going to commit to getting this landed in a 2.0 line this week for all involved here.

HoldYourWaffle:
Thank you for (temporarily) reopening this.
I just wanted to say that I honestly never considered merging this into DT. I prefer to include (fixed) typings in modules because DT has its fair share of issues (which - among other things - caused this mess in the first place).

But you're right, in this case it may be the proper solution. After all, these situations are sort of what DT was meant for in the first place.

I'll see if I can create a PR over at DT soon, if you chip in to say that these typings match (like you did when you reviewed the PR) it *should* go smoothly. I'd still prefer including the typings in the module itself, but this way you don't have to rush v2.

dougwilson:
Sounds good. I will be happy to confirm the types on there.  

Note I only temporary locked it; it will remain open unless there is a reason to lock again, so you don't need to worry about it just suddenly going locked without warning :)

HoldYourWaffle:
I have just created [the PR](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/41809) to update the definitions over at DT.

HoldYourWaffle:
I have just created a [second-attempt PR](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/46576) to update the definitions over at DT, could you (@dougwilson) check if these are still accurate after a year?

HoldYourWaffle:
I have updated the imported definitions to the current latest DT candidate ([PR](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/46576)). I'll update my [published fork](https://www.npmjs.com/package/@holdyourwaffle/express-session) shortly.

To increase my developer-citizen-karma, I have copy-pasted (and reformatted) the documentation from the README into the type definitions as tsdoc. Unfortunately I couldn't find anything on `CookieOptions.signed`, `CookieOptions.encode`, `Store.regenerate`, `Store.load` and `Store.createSession`.

Did I miss something or is there really documentation on these properties/methods?

HoldYourWaffle:
I'm happy to inform you that https://github.com/DefinitelyTyped/DefinitelyTyped/pull/46576 was merged!

I'd still like these typings to be merged here some day, because (as you might've seen) DT can be a bureaucratic mess...

PR #664, opened 2019-05-30
originalMaxAge should always be restored.

jbialobr:
If `expires` is a Date, then originalMaxAge is not restored and session is not being rolled.

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,6 +1,7 @@
 unreleased
 ==========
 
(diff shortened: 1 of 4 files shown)


jbialobr:
Tests added. They pass unchanged for the 1.15.6 version.

dougwilson:
Thank you. This is a fix for a regression from the commit cc5c83abf710c6e9e4d0bb887777b44faefd2251 . The test will make sure we don't regress this in the future 👍

PR #672, opened 2019-07-10
bump cookie and safe buffer

knoxcard:
bump cookie and safe buffer


dougwilson:
Another user created a PR for the cookie update (#674). I will pull that out of this one and leave it just as the `safe-buffer` change 👍

PR #677, opened 2019-08-06
Made "rolling" option better understandable for beginners

BorntraegerMarc:


README.md
```diff
@@ -228,8 +228,8 @@ likely need `resave: true`.
 ##### rolling
 
 Force a session identifier cookie to be set on every response. The expiration
-is reset to the original [`maxAge`](#cookiemaxage), resetting the expiration
-countdown.
+is reset to the original [`maxAge`](#cookiemaxage) on every request, resetting the expiration
```
dougwilson: I'm not sure what the addition of the "on every request" here means; at least there is nothing I know of that would touch this during the request in this module, just the response. Can you clarify?

README.md
```diff
@@ -228,8 +228,8 @@ likely need `resave: true`.
...
+countdown. This leads to sessions being "renewed": Can be useful if you want to keep users logged in while they are active, so they don't need to re-authenticate.
```
dougwilson: I think there is some confusion here; the rolling only applies to the cookie, which is just a session ID link. The actual session itself is always renewed even without this rolling option, though that can depend on which store module you're using, as they can vary in their behavior.

README.md
BorntraegerMarc: Right, so actually I meant on every response. I was just looking at implementation from my express app and got a bit confused. adapted text

README.md
BorntraegerMarc: Absolutely. Adapted the terminology in my latest commit so it makes more sense based on your feedback. As a beginner it's a bit hard to get the terminology right but now I understand what you mean with "session ID link"

BorntraegerMarc:
requested changes adapted

README.md
dougwilson: Thank you. I will do some re-wording here as well, especially around users logged in, as this module has no concept of users and logging in; a session can contain any data for any reason, so I'll put it in the context of the general data storage that this module is enabling.

Issue #633, opened 2019-02-18
Express-Session not persisting / Working in production even having a store option in the session options


anikethsaha:
I saw many post regarding this issue but couldn't get the answer.
In many posts the answers are to have a session storage.

In my project I am having a session - storage but still the problem
persist

I am using connect-session-firebase and Firestore for storing the session but still I am not getting a persistent session. 

It's working fine in development but not in Production even having a
sessionstorage

My codes
server.js

    
    const session = require('express-session')
    const cors = require('cors')
    const helmet = require('helmet')
    const { port , sessionSecretKey} = require('./configs/config')
    var cookieParser = require('cookie-parser')
    var flash = require('connect-flash');

    const { store} = require('./configs/sessionStorage/firebaseSessionStorage')
    const {dbname,MONGODB_URL,sessionKeys} = require('./configs/config.js')
    //database connection
    mongoose.connect(MONGODB_URL,{
      useNewUrlParser: true
    });
    const app = express();

    app.use(helmet());

    app.use(bodyParser.urlencoded({extended: true, limit: '50mb'}))
    app.use(bodyParser.json({limit: '50mb'}))
    
    app.use(cookieParser());
    
    app.use(session({
      store,
      secret: sessionSecretKey,
      resave: true,
      saveUninitialized: true,
      cookie: {
        secure: process.env.NODE_ENV == "production" ? true : false ,
        maxAge: 1000 * 60 * 60 * 24 * 7
      }
    }));

Now I have created a middleware with every request just to count the view using session

    app.use(function(req,res,next){
    
      res.locals.user = req.user || null;
      if(req.session.views){
        req.session.views += 1
        req.session.save();
      }else{
        req.session.views = 1
        req.session.save();
      }
      console.log('req.session.views', req.session.views)
      next();
    })

Its Giving output `req.session.views 1` in production

PM2 script for start
`pm2-runtime start ecosystem.config.js --env production`

ecosystem.config.js

    module.exports = {
      apps : [{
        name: 'server',
        script: 'server.js',
        // Options reference: https://pm2.io/doc/en/runtime/reference/ecosystem-file/
        instances: 2,
        autorestart: true,
        watch: false,
        max_memory_restart: '1G',
        env: {
          NODE_ENV: 'development'
        },
        env_production: {
          NODE_ENV: 'production'
        }
      }],
    
      deploy : {
        production : {
          user : 'node',
          host : '212.83.163.1',
          ref  : 'origin/master',
[...]

PR #708, opened 2019-11-04
src: more debug statements

gireeshpunathil:
Add debug statement (enabled only in debug mode) at vital session life cycle control flow points

Refs: https://github.com/expressjs/session/issues/633#issuecomment-471110093

Generally improves ability to diagnose issues.

(from a high level view on the outstanding issues in this repo, it is evident that the most prominent type is diagnosing session persistence across client requests, so this is a small step towards helping there)

### DIFF
--- a/index.js
+++ b/index.js
@@ -179,6 +179,7 @@ function session(options) {
   return function session(req, res, next) {
     // self-awareness
     if (req.session) {
+      debug('re-using existing session')
       next()
       return
     }
@@ -302,13 +303,13 @@ function session(options) {
 
       if (shouldDestroy(req)) {
         // destroy session
-        debug('destroying');
+        debug('destroying session store');
         store.destroy(req.sessionID, function ondestroy(err) {
           if (err) {
             defer(next, err);
           }
 
-          debug('destroyed');
+          debug('session store destroyed');
           writeend();
         });
 
@@ -333,6 +334,7 @@ function session(options) {
             defer(next, err);
           }
 
+          debug('session saved');
           writeend();
         });
 
@@ -357,6 +359,7 @@ function session(options) {
 
     // generate the session
     function generate() {
+      debug('session generate');
       store.generate(req);
       originalId = req.sessionID;
       originalHash = hash(req.session);


index.js
```diff
@@ -302,13 +303,13 @@ function session(options) {
 
       if (shouldDestroy(req)) {
         // destroy session
-        debug('destroying');
+        debug('destroying session store');
```
dougwilson, line 306: This is destroying the session, not the session store

index.js
```diff
@@ -302,13 +303,13 @@ function session(options) {
...
         store.destroy(req.sessionID, function ondestroy(err) {
           if (err) {
             defer(next, err);
           }
 
-          debug('destroyed');
+          debug('session store destroyed');
```
dougwilson, line 312: this destroys the session, not the store

index.js
```diff
@@ -333,6 +334,7 @@ function session(options) {
             defer(next, err);
           }
 
+          debug('session saved');
```
dougwilson, line 337: this should probably be placed in conjunction with the saving log, as the session can be saved from many different places

index.js
gireeshpunathil: @dougwilson - addressed, changed the log message

index.js
gireeshpunathil: ditto as above

index.js
gireeshpunathil: done, moved to `save` method.

index.js
```diff
@@ -393,6 +395,7 @@ function session(options) {
         debug('saving %s', this.id);
         savedHash = hash(this);
         _save.apply(this, arguments);
+        debug('session saved');
```
dougwilson, line 398: This prints before the session is saved, as the saving operation is async.

dougwilson: This is based on I am assuming that this line is desired to be similar to the destroy logs, with the behavior of "destroying.." being logged before the store call and then "... destroyed" called when the store returns back to this module without an error.

gireeshpunathil: thanks @dougwilson . yes, the desire is to follow `destroy`. the intent is simple - many issues in this repo are about session not being saved or inability to know about it.

thinking about it more, I can achieve this at the `save` site only via structural changes to the `save` function. that is, anchoring the `save` call with a custom callback, check for errors, or else print the debug message and then invoke the original callback, if any.

Before implementing it and submitting for your review, I want to pro-actively check with you on it - will you be open for such a change?

dougwilson: Hi @gireeshpunathil yes, that sounds good to me, to just mimic the destroy way in the save area.

gireeshpunathil: @dougwilson - done, pls have a look. Here is a sample output from a simple program with debug on, for your reference:
```
  express-session fetching qs3pPenGDQy0hOXWCw_G8lD_ZLJw5AA1 +0ms
  express-session no session found +5ms
  express-session session generate +1ms
  express-session saving 7G28RiMV20yaNOzU3UeC5_qJMKek7X6f +4ms
  express-session set-cookie connect.sid=s%XXX; Path=/; Expires=YYY; HttpOnly +4ms
  express-session session saved +3ms
```

index.js
```diff
@@ -389,10 +391,14 @@ function session(options) {
...
 
-      function save() {
+      function save(cb) {
         debug('saving %s', this.id);
         savedHash = hash(this);
-        _save.apply(this, arguments);
+        _save.call(this, function(err) {
+          if(!err)
+            debug('session saved');
+          cb(err);
```
dougwilson, line 400:
```suggestion
          if (error) {
            return callback(error)
          }

          debug('session saved')
          callback.apply(this, arguments)
```

gireeshpunathil:
- I have addressed the review comments
 - I have added the word session in the debug statements in touch code
 - request to have a look
 - request to remove the low priority tag

dougwilson:
> request to remove the low priority tag

For what reason? Do you think this is not a low priority against the other aspects the team is attempting to move forward like Express 5.0? I'm trying to apply tags around priorities to help people focus the appropriate attention to get larger efforts like 5.0 moving forward. IMO adding some extra debug wording seems low priority to me, compared to things like new features, bug fixes, etc.

The tag itself doesn't mean that it gets no attention, just that it is comparatively low priority. I'm happy to remove it, but just unsure on the reasoning for removing it, as you didn't provide any.

index.js
```diff
@@ -339,13 +340,13 @@ function session(options) {
         return writetop();
       } else if (storeImplementsTouch && shouldTouch(req)) {
         // store implements touch method
-        debug('touching');
+        debug('touching the session')
```
dougwilson, line 343: Any reason this ones contains `"the"` while the others (like `destroying session`, `generating session`, etc.) does not? Should all the others contain "the" or should this one remove the "the"?

gireeshpunathil: fixed it, removed `the`

index.js
```diff
@@ -389,10 +391,17 @@ function session(options) {
...
       }
 
-      function save() {
+      function save(callback) {
         debug('saving %s', this.id);
         savedHash = hash(this);
-        _save.apply(this, arguments);
+        _save.call(this, function (err) {
+          if (err) {
+           return callback(err)
```
dougwilson, line 399:
```suggestion
            return callback(err)
```

Sorry, doing it through the GH UI is hard to count spaces 😆

gireeshpunathil: thanks, fixed it.

gireeshpunathil:
I am not arguing with you; but I cannot find a reason why this  is a low priority: the reasoning for this change is provided in the first comment - to improve self diagnosis of a class of issues that report malfunction of session life-cycle events, more specifically users suspecting missing session persistence. Plus, there are 27 other PRs in this repo, and none of those are low pririty items!

But I am defenitely not arguing with you nor questing your judgement. Also, if this is to compete with express 5.0 for project's attention and focus, I will prefer the later, as that has more user demand than the former.

dougwilson:
I already provided the reason why it is low priority above :) The tag has nothing to do with the timeline for landing the change; it is simply to provide a category of the change itself. The argument that the others are not tagged right is not an argument this is not a low priority change; getting things tagged correctly is actually your primary function as a triager :) So if other PRs are not tagged correctly, it sounds like that is an issue that needs t be addressed with at least the triagers; I was simply tagging this PR since I happened to just interact with it. If it is more important for me to go through all the issues and PRs in this repository and get them all tagged appropriately vs working on other items, I can certainly do that, but the idea was that we're bringing on the large group of triagers to perform these kinds of tasks such that folks like myself can actually focus on the larger-at-hand items.

gireeshpunathil:
> I already provided the reason why it is low priority above :)

I have given response to that - it is not some random debug statements; it is derived from an insight from this repo's issue backlog.

> The argument that the others are not tagged right is not an argument this is not a low priority change; getting things tagged correctly is actually your primary function as a triager :)

There are PRs which are 6 years old with no activities for as long; and acting on those in a meaningful and timely manner is actually your primary function as a maintainer :)

But then let us agree that we are not here to remind each other on their responsibilities.

> So if other PRs are not tagged correctly, it sounds like that is an issue that needs t be addressed with at least the triagers; I was simply tagging this PR since I happened to just interact with it.

It looks incorrect approach to me. If you just happened to interact with a work item cannot be used to derive a priority decision? Priorities are always relative; without looking at other PRs, a 'low priority' item cannot stand in itself. Low against what?

dougwilson:
> I have given response to that - it is not some random debug statements; it is derived from an insight from this repo's issue backlog.

I get it, but we're also not even actively telling folks to use what is there when triaging issues (see recent issues like #763 )

I think that this conversation is getting a bit out of hand and off topic, and frankly into a bit of the offensive realm. If you could, I would suggest if you could please step away for a bit. I'm going to lock this PR for now.

dougwilson:
@gireeshpunathil I think at this point going forward, I just can no longer act on any issues and PR you raise. You will need to get something else involved to move forward with your issues and PRs, as I do not feel comfortable in interactions between us.

PR #713, opened 2019-11-13
allow tolerance when checking cookie's maxAge

gabylb:
#### Checklist
- [x] npm install && npm test passes
- [x] commit message follows [commit guidelines](https://github.com/nodejs/node/blob/master/doc/guides/contributing/pull-requests.md#commit-message-guidelines)

#### Description of change
- test/cookie.js fails intermittently `(59999 !== 60000)` when it checks the cookie's 'maxAge' against the value used when the cookie was created. Since this property is 'live', its current value (in ms) can occasionally decrease by the time the check is made. The test already deals with the (also live) 'expires' property in a similar way.

### DIFF
--- a/test/cookie.js
+++ b/test/cookie.js
@@ -91,7 +91,9 @@ describe('new Cookie()', function () {
         var maxAge = 60000
         var cookie = new Cookie({ maxAge: maxAge })
 
-        assert.strictEqual(cookie.maxAge, maxAge)
+        assert.strictEqual(typeof cookie.maxAge, 'number')
+        assert.ok(cookie.maxAge - 1000 <= maxAge)
+        assert.ok(cookie.maxAge + 1000 >= maxAge)
       })
 
       it('should accept Date object', function () {


test/cookie.js
```diff
@@ -91,7 +91,8 @@ describe('new Cookie()', function () {
         var maxAge = 60000
         var cookie = new Cookie({ maxAge: maxAge })
 
-        assert.strictEqual(cookie.maxAge, maxAge)
+        assert.ok(cookie.maxAge <= maxAge);
```
dougwilson: This check seems wrong. For example, it still passes even when `cookie.maxAge` is `undefined`, which doesn't seem in the spirit of validating that the maxage was actually set.

gabylb: In my test, that statement does fail as expected If `cookie.maxAge` was undefined.
I tested with this code:
```
        cookie.maxAge = undefined;
        assert.ok(cookie.maxAge <= maxAge);
```
and got:
```
      AssertionError [ERR_ASSERTION]: The expression evaluated to a falsy value:
  assert.ok(cookie.maxAge <= maxAge)
```
Anyway I've added an extra check.

gabylb:
@dougwilson in the 'coverage decreased' details link, it's not clear to me how/why my change decreased coverage; can you please help clarify.

PR #715, opened 2019-11-17
fix: reload calls exceeding stack track in case of errors

Zirak:
When `reload`'s callback is called, it invokes `wrapmethods` on the
request session, ensuring the new session's `reload` call is overriden
with a custom `reload`, calling the original. However, there's no
guarantee a new session is created, resulting in `reload` overriding
an overriden `reload`! Given enough such override-overrides, the stack
will overflow, gruesomely erupting into several hundreds of lines of
callstack, leaving the lonely developer to wonder how he has reached
such a mess, reflecting on several important life choices.

Note that the testing framework, for reasons beyond my current
comprehension, did not correctly act upon this `RangeError`. When
running the new test case without the fix, the `mocha` call hangs
seemingly indefinitely. That may be deserving of further research, but
most deserving of a word of warning.

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,3 +1,8 @@
+unreleased
+==========
+
+  * Fix internal method wrapping error on failed reloads
+
 1.17.0 / 2019-10-10
 ===================
 
--- a/index.js
+++ b/index.js
@@ -376,6 +376,16 @@ function session(options) {
       wrapmethods(req.session)
     }
 
+    function rewrapmethods (sess, callback) {
+      return function () {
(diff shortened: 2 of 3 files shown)


ghinks:
**thank you** @Zirak. We have some more developers looking into the express-session module.  Let me  take a look at your PR. Are you still interested in supporting your change it has been several months since you made the PR?

Zirak:
Hey hey, yes I'm still invested in this PR.
Thanks for taking a look!

Zirak:
Thanks!

Issue #710, opened 2019-11-06
Session data is not being stored (Inconsistent)


deveshMantra:
Trying to save extra data in session and redirecting to the other page after coming back the extra data is not stored rest of the thing in session is available this is happening inconsistently.

server.js file
```
```app.use(cookieParser())
app.use(helmet())
app.use(session({
  secret: 'some secret key',
  resave: false,
  saveUninitialized: false,
  store: memoryStore
}))```
```

once the user logs in the system we are storing the following data in session.
```
```{ cookie: { originalMaxAge: null, expires: null, httpOnly: true, path: '/' },
  auth_redirect_uri: 'http://localhost:3000/resources?auth_callback=1',
  'keycloak-token': '{"access_token":"someToken","expires_in":86400,"refresh_expires_in":43200,"refresh_token":"someToken","not-before-policy":0,"session_state":"89e0e02e-1414-4043-ba1b-a61f23e5dfa4"}',
  userId: '2752-4e2e-bb6d-d6693c3190d5',
} ```

```

User tries another api we are storing more data into the existing session.

` mergeAccountInfo: 
   { initiatorAccountDetails: 
      { userId: '9b7de845-2752-4e2e-bb6d-d6693c3190d5',
        sessionToken: 'some token',
        redirectUri: '/resources',
        test: 'mockValue' } }`

To save the session below is code I am trying to use 

```
req.session.newDataToSave = {'some data': ' some more data'}
   req.session.save(function (result) {
    const url = redirect to different url;
    res.redirect(url)
    });
```

after i come back to the redirect url i see the new data that we just saved is not present while the other data is present that was saved after user logged in to system.

Any help on same is highly appreciable @beeman @voxpelli @mikeal @juliangruber @tj

HarshithaKP:
@deveshMantra I have written a simple example to demonstrate session persistence with redirection here : https://gist.github.com/HarshithaKP/3d9ad81138245aa7527bf385d37daba7 
It is working fine for me, let me know if it helps you.

gireeshpunathil:
@deveshMantra - was your issue resolved with the example?

deveshMantra:
@gireeshpunathil 
The issue is resolved in my code I did a hack for the same as it was production issue but I think we need to find another solution in the package itself to avoid such conditions.
There was a race condition due to which session was being overwritten.
Here is what was happening.
req1 and req2 both were coming at the same time. 
req1 - when request 1 comes it fetches the session data(say A) it updates session(say B).
[...]

Issue #660, opened 2019-05-07
Session not saving when redirect


alex55132:
When an user logs in the data is saved in `req.session`, then I redirect the user to the homepage using `res.redirect(url);`. The problem appears when that redirect execute, the data from the session get lost but if instead of redirecting I go to the homepage manually (with the browser), the session data is available.

Here are some snippets of my code:

main.js:
```
var session = require('express-session');
var mySQLStore = require('express-mysql-session')(session);

let sessionConfiguration = session({ /*Session configuration */});

function checkSession(req, res, next) {
  if(req.session.user) {
    console.log("Exists");
    console.log(req.session.test);
    next();
  } else {
    console.log("No exist :( ");
    next();
  }
}

app.use('/', sessionConfiguration, checkSession, router);

```

redirect.js:

```
router.get('/route', function (req, res) {  

     req.session.test = "hey"
     console.log(req.session.test);

      //Data get lost here
      res.redirect("/");

  });
```

Also, the data of the session that remains when the redirect is done is this:

`{"cookie":{"originalMaxAge":3600000,
"expires":"2019-05-07T21:15:06.727Z",
"httpOnly":false,
"path":"/",
"sameSite":true}}`

dougwilson:
Due to how most web browser handle redirects, you need to save before the direct (https://github.com/expressjs/session#sessionsavecallback).

alex55132:
@dougwilson Also tried that, still not working

dougwilson:
What does your code look like when trying it?

alex55132:
@dougwilson In redirect.js:

```
router.get('/route', function (req, res) {  

     req.session.test = "hey"
     console.log(req.session.test);

      req.session.save(function(err) {
         if(!err) {
             //Data get lost here
             res.redirect("/");
         }
      });

  });
```

dougwilson:
Hm, that should work. I'll reopen so we can investigate. Can you provide a full app and reproduction steps?

alex55132:
@dougwilson Well the full app is the code provided earlier with the express module and the line which listens for connections, its only that

dougwilson:
Right but it doesn't run as is. I modified it enough to actually run and then I had no issue and session didn't loose data after I made the session save modification as well.

alex55132:
@dougwilson Can you show the code please? Which version are you using? This is really really weird :/

dougwilson:
I am using latest. I already left work and you'll have to wait until I get back on Thursday to get the code I had.

[...]

PR #740, opened 2020-04-06
Update README.md

tranvansang:
Remove `redirects` from examples of when `session.save()` should be called.

Session is saved even after a redirect.

### DIFF
--- a/README.md
+++ b/README.md
@@ -359,8 +359,7 @@ session data has been altered (though this behavior can be altered with various
 options in the middleware constructor). Because of this, typically this method
 does not need to be called.
 
-There are some cases where it is useful to call this method, for example,
-redirects, long-lived requests or in WebSockets.
+There are some cases where it is useful to call this method, for example, long-lived requests or in WebSockets.
 
 ```js
 req.session.save(function(err) {


dougwilson:
Yes, they are, but users have continuously reported that it is necessary to wait for save then redirect.

tranvansang:
> Yes, they are, but users have continuously reported that it is necessary for their environment to wait for save then redirect

Then, why don't we figure out exactly when session.save() is required?
Isn't the current docs ambiguous about this?

dougwilson:
Yes, we have spent a lot of time diagnosing, and it just ends up no where, as it's no reproducible on our end, and the users are not familiar with the code base.

> Isn't the current docs ambiguous about this?

I don't think so. The docs state the the session is automatically saved on every response (a redirect response is still a response). The save method part you changed just generically lists some "useful" cases where using the save method could come in handy--you're free to use or not use it as your application calls for it.

If you can lend a hand to help with the save/redirect problem, there are numerous open issues about it like #710 #660 etc.

tranvansang:
One of the reason is explained here

https://gist.github.com/HarshithaKP/3d9ad81138245aa7527bf385d37daba7#file-sessionpersistence-js-L21

>     // While sessions with memoryStore may work fine with multiple requests/redirects, external
>     // stores can cause race conditions - between next client request and the store, depending
>     // on who runs faster. So save the session explicitly, prior to the response/redirect.

dougwilson:
So I use the popular Redis one mostly, and never had such an issue at least that I've experienced. I'm also unsure if that is the case, what makes it specific to `res.redirect` vs any other type of response? The comment from that Gist I can't tell if it is saying it is specific to redirects, or just a generalized statement.

tranvansang:
I think it is very general to any type of response as long as the store is asynchronous.

It is not hard to reproduce the behaviour. I tried edit session/memory.js and successfully reproduced

```
MemoryStore.prototype.set = function set(sessionId, session, callback) {
  setTimeout(() => {
    this.sessions[sessionId] = JSON.stringify(session)
    callback && defer(callback)
  }, 5000)
}
```


I also found one thing that the browser (latest chrome/ubuntu 18.04), does redirect before the response ends.

I got the log from the redirect response before the redirector finishes.

PR #744, opened 2020-04-22
add getDomainFromRequest option, for when domain needs to be calculated

paulfitz:
Thanks for express-session!  Here's a small addition that is perhaps a bit niche but has been useful for us.

This adds a `getDomainFromRequest` option which allows `cookie.domain` to be computed from the request rather than set as a constant.

As an example of when this is useful, imagine serving a single app from `*.foo.com` and `*.bar.com`, with the cookie domain being `.foo.com` for the `*.foo.com` domains and `.bar.com` for the `*.bar.com` domains.

This may be related to https://github.com/expressjs/session/pull/311

### DIFF
--- a/README.md
+++ b/README.md
@@ -182,6 +182,18 @@ app.use(session({
 }))
 ```
 
+##### getDomainFromRequest
+
+Optional function to override the `Domain` `Set-Cookie` attribute.
+Provide a function that returns a string that will be used as the
+cookie domain. The function is given `req` as the first argument if
+you want to use some value attached to `req` when generating the
+domain (as for `genid`).  This can be useful if the same app is
+accessed from different families of domains, some of which should
+share cookies.
+
+If this function is not provided, `cookie.domain` is used instead.
+
 ##### name
 
 The name of the session ID cookie to set in the response (and read from in the
--- a/index.js
+++ b/index.js
@@ -114,8 +114,15 @@ function session(options) {
   // get the cookie signing secret
   var secret = opts.secret
 
+  // get a function for computing the domain from the request
+  var getDomainFromRequest = opts.getDomainFromRequest
+
   if (typeof generateId !== 'function') {
     throw new TypeError('genid option must be a function');
+  }
+
+  if (getDomainFromRequest !== undefined && typeof getDomainFromRequest !== 'function') {
+    throw new TypeError('getDomainFromRequest option must be a function');
   }
 
   if (resaveSession === undefined) {
@@ -159,6 +166,12 @@ function session(options) {
     req.sessionID = generateId(req);
     req.session = new Session(req);
     req.session.cookie = new Cookie(cookieOptions);
+
+    if (getDomainFromRequest) {
+      // If a function was specified for computing the domain
+      // from the request, then use it now.
+      req.session.cookie.domain = getDomainFromRequest(req);
+    }
 
     if (cookieOptions.secure === 'auto') {
       req.session.cookie.secure = issecure(req, trustProxy);
--- a/test/session.js
+++ b/test/session.js
@@ -891,6 +891,37 @@ describe('session()', function(){
     });
   });
 
+  describe('getDomainFromRequest option', function(){
+    it('should reject non-function values', function(){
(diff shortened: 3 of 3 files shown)


dougwilson:
It seems your scenario is different from that one, and seems achievable though the use of middleware, which is what makes Express so flexible. Have you considered something like
```js
app.use(vhost('*.foo.com', session(fooConfig)))
app.use(vhost('*.bar.com', session(barConfig)))
```

You can pass the same store to both if you want to store sessions in the same place.

paulfitz:
Thanks for the suggestion! I did consider something similar. In my actual case, the specific domain families are not available to the code, so the middleware needed to do a bit more gymnastics. Overall it felt clumsy, and cleaner to solve in express-session, leaving me with a single session object. However, I can see this is unlikely to be a common case, so it is fine with me if you want to close this out, I won't be offended :-)

dougwilson:
It's no problem, and I'm not rejecting it to be clear :) I'm more trying to better understand the use-case so we can get to a situation where this is generically solvable and doesn't end up where this needs to be done for every other option, and then someone needs it to be async, etc. etc. :)

There may be a clean way, but I guess I would say it's hard to work together without understanding what you're trying to accomplish. I believed I understood from your initial issue, but it sounds now that what was written there doesn't actually represent your true goals to get from this change.

Is it possible to outline them?

paulfitz:
I'd now edit:

> As an example of when this is useful, imagine serving a single app from `*.foo.com` and `*.bar.com`, with the cookie domain being `.foo.com` for the `*.foo.com` domains and `.bar.com` for the `*.bar.com` domains.

to be: imagine serving a single app from several domain families (`*.foo.com`, `*.bar.com`, etc.) where the cookie domain should omit the subdomain, and the set of domain families is determined outside of the code (e.g. in dns and load balancer configuration).

Does that make any more sense?  Thanks for your patience.

dougwilson:
Hi @paulfitz thank you. That does make sense. Sounds like you could use `vhost` to accomplish this (I just wrote it up and didn't extensively test it). The nice thing is that this works with _any_ middleware config, and not even just this module. If this is a pattern that is useful, having a module on npm to wrap around middlewares would be super useful to everyone too.

```js
// example lib that exports a middleware to use for your use-case:
var session = require('express-session')
var vhost = require('vhost')

module.exports = function (options) {
  var cache = Object.create(null)
  return vhost(/^(?:.+\.)?([^.]+\.[^.]+)$/, function (req, res, next) {
    var domain = req.vhost[0]
    var mw = cache[domain]

    if (!mw) {
      var opts = Object.assign({}, options)
      opts.cookie = Object.assign({}, opts.cookie || {})
      opts.cookie.domain = '.' + domain
      cache[domain] = mw = session(opts)
    }

    mw(req, res, next)
  })
}
```

ramiroaisen:
Upvote for this I have the same trouble here.
I want to share sessions between subdomains,
it would be really nice to have a callback domain config instead of a constant
something like this
```javascript
 session({
  domain: (req: IncommingMessage): string | undefined => req.headers.host // or whatever
})
```
Note that this signature is fully backwards compatible as ```session({domain: "example.com"})``` still works as expected

uriesk:
Got the same usecase now.
I do not like the vhost or custom middleware solutions because they all create a new session middleware at every single request.
I would prefer callback function that allows to change any cookie settings in an already created session middleware (maybe someone wants to dynamically adjust the expire time?).

dsagal:
The implementation in this PR is simple, clean, and well-documented, and addresses nicely the use-cases where the domains of the app are dynamic, without affecting behavior when the new option isn't used. Any reason not to merge?

bjohansebas (approved):
No strong opinions on this functionality, if it helps people then I'm fine with adding it.

Goryounov:
Hello! I really need this feature, please merge it.

bjohansebas (changes requested):
I think allowing the configuration to accept a function, as done in https://github.com/expressjs/session/pull/1027, would solve the problem here and we wouldn’t need to add a new option. wdyt?

PR #746, opened 2020-05-05
test: writeHead should merge

ghinks:
The test checking that cookies should be preserved before writeHead is called should check in addition that when writeHead is called then the merge that is documented in the nodeJS docs does in fact get carried out in the manner we expect it to.

**writeHead** should merge on top of the existing value.

### DIFF
--- a/test/session.js
+++ b/test/session.js
@@ -1892,6 +1892,21 @@ describe('session()', function(){
           .expect(shouldSetCookieToValue('previous', 'cookieValue'))
           .expect(200, done)
         })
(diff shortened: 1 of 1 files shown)


ghinks:
@dougwilson .agreed, I'll update it.

dougwilson:
Hey @ghinks I see you marked it as a draft. Is it OK if I unmark it as a draft, or are you looking to do something more on it before merge review?

ghinks:
Pls unmark

Issue #750, opened 2020-05-18
Docs should describe how to securely generate session secret


pkrasicki:
Docs should describe how to securely generate session secret and why it's important. I'm mentioning this again, because my other issue #734 was closed, but the problem still remains.

ghinks:
Thank you for the continued discussion. In general a **secure** systems would in fact not hard code the secret in the javascript. It would be an environment variable(s) and would not be visible within the repository within which express exists. 

The **phrase securely generate session secret** really does not mean how do I generate a secret but how do I **keep the secret**. The secret itself being a text string that is not easy to read or remember.

Keeping the secret would involve making sure it was out of the code base, and would involve changing it periodically.

This is actually security advice. 

I will make a PR but it is up to you to collaborate with us to determine the correct language.

ghinks:
https://github.com/expressjs/session/pull/752 has been raised. I would encourage @pkrasicki to review and give feedback

ghinks:
closing as PR 752 has been raised.

PR #752, opened 2020-05-19
docs: session secret advice

ghinks:
Add some general advice on using
- decent secret key
- storing key out of code
- updating key

referencing the request for some clarification on secret documentation from [issue 750](https://github.com/expressjs/session/issues/750#issue-620308366) this draft PR has been raised for review.

### DIFF
--- a/README.md
+++ b/README.md
@@ -272,7 +272,22 @@ it to be saved. *This has been fixed in PassportJS 0.3.0*
 This is the secret used to sign the session ID cookie. This can be either a string
 for a single secret, or an array of multiple secrets. If an array of secrets is
 provided, only the first element will be used to sign the session ID cookie, while
-all the elements will be considered when verifying the signature in requests.
+all the elements will be considered when verifying the signature in requests. The
+secret itself should be not easily parsed by a human and would best be a random set
+of characters. A best practice may include:
+
+  - The use of environment variables to store the secret, ensuring the secret itself
+    does not exist in your repository.
+  - Periodic updates of the secret, while ensuring the previous secret is in the
+    array.
+
+Using a secret that cannot be guessed will reduce the ability to hijack a session to
+only guessing the session ID (as determined by the `genid` option).
+
+Changing the secret value will invalidate all existing sessions. In order to rotate
+the secret without invalidating sessions, provide an array of secrets, with the new
+secret as first element of the array, and including previous secrets as the later
+elements.
 
 ##### store
 


pkrasicki:
This looks good!

pkrasicki:
What do you think of adding a small warning about what could happen if the secret is not random enough?

ghinks:
@pkrasicki additional comments have been added.

pkrasicki:
This is great!

ghinks:
I believe making a note about changes to the key would help and will update the PR.

dougwilson:
Hey @ghinks @jonchurch I pushed up (as it's own commit to make it easier to see and/or revert) some changes I thought would be useful. Most of it was just "word smithing", but I also provided (imo) a clearer explanation of what guessing a secret would enable. PTAL.

README.md
```diff
@@ -272,7 +272,22 @@ it to be saved. *This has been fixed in PassportJS 0.3.0*
...
+all the elements will be considered when verifying the signature in requests. The
+secret itself should be not easily parsed by a human and would best be a random set
+of characters. A best practice may include:
+
+  - The use of environment variables to store the secret, ensuring the secret itself
+    does not exist in your repository.
+  - Periodic updates of the secret, while ensuring the previous secret is in the
+    array.
+
+Using a secret that can be guessed will reduce the ability to hijack a session to
```
ghinks, line 284: will change from **can** to **cannot**

dougwilson: No, can js what I mean. Changing it to cannot is wrong, as then it will say if you cannot guess the secret, then you can hijack the session with only guessing the session ID.

ghinks:
I made one small grammar change. 

Using a secret that **can** be guessed will reduce the ability

to 

Using a secret that **cannot** be guessed will reduce the ability

dougwilson:
No, can is what I mean. Changing it to cannot is wrong, as then it will say if you cannot guess the secret, then you can hijack the session with only guessing the session ID.

dougwilson:
The purpose of having the secret is that if you don't know the secret, you have to guess both the secret and the session ID, but if you know the secret, then you only have to guess the session ID in order to hijack a session.

ghinks:
ok, re-reading it, you are correct. apologies

dougwilson:
Thanks @ghinks ! I left it for a bit to settle and I think that this PR has indeed settled down, so let's get it 🚢 ! If there are further things that turn up about this after the fact, we can follow it up with another PR of course.

README.md
0xnoob: Hey, it looks like this was never actually changed? I can't really view the entire file (`Commit range not found`), only the lines in the preview. 
The current phrasing led me to believe the signature doesn't prevent from guessing the session ID (if you consider the signature not part of the session ID), but from something else that isn't mentioned. which made me [asking on stackoverflow](https://stackoverflow.com/questions/64667592/what-is-the-purpose-of-the-secret-in-express-session-middleware).  
My solution would be the following phrasing, but I'm not a native english speaker (as you probably figured out):
>Using a secret that **cannot** be guessed will reduce the ability to hijack a session **by** only guessing the session ID (as determined by the genid option).

PR #764, opened 2020-07-02
docs: add yarn instructions to install express-session

Pardim93:


dougwilson:
Thank you for that, but in general for the adding yarn install PRs, we just close them. The reason is that the only reason we have the npm install command is for users who had no clue how to install packages set (those new to the ecosystem is the target audience). The idea is that those who are using yarn are outside of that audience and can see a npm install command and know the related yarn (or pnpm, or insert other package manager here) command to run from that.

PR #767, opened 2020-07-08
Fix res.end patch to always commit headers

### LINKED ISSUE #766: Sending after request end doesn't throw error, hangs response instead
Hi, I've noticed that when including `express-session`, the `res.end` behavior (from what I'm assuming is due to the proxied [res.end](https://github.com/expressjs/session/blob/master/index.js#L247)) deviates from what I would expect in handling errors. 

Generally if I were to call `res.end` and then later inadvertently call `res.send` after that (perhaps in another middleware handler), I would receive an `ERR_HTTP_HEADERS_SENT`. That's what I was expecting to see—however, instead, if I include the session middleware, the headers send but it waits for 5s (which I think is the default keepAlive timeout) until erroring out with a content length mismatch (since it is setting `Content-Length` but the body doesn't go through). In my opinion, it would be helpful to have that original error bubble through if possible, since it was not clear that including this middleware would cause this API change to `res.end` as a side effect.

```js
var express = require("express");
var app = express();

// Uncommenting the middleware usage leads to different functionality in handling the duplicate response sending
app.use(
  require("express-session")({
    secret: "keyboard cat",
    resave: true,
    saveUninitialized: true,
  })
);

app.get("/", function (req, res) {
  res.end();
  // Doesn't throw error! Waits for keepAlive duration
  res.send("This is a test!");
});

app.listen(5000);
```

dougwilson:
This adds a missing command to commit the progressive headers within `res.end` before attempting to read them. This has the standard side-effect of ensuring a second call to `res.end` will fail with a headers already set error. This is the same behavior other modules like `compression` does to get the headers set, same as the internal Node.js code of `res.end`.

Fixes #766

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,6 +1,7 @@
 unreleased
 ==========
 
+  * Fix `res.end` patch to always commit headers
   * deps: cookie@0.4.1
   * deps: safe-buffer@5.2.1
 
--- a/index.js
+++ b/index.js
@@ -272,6 +272,10 @@ function session(options) {
           return ret;
         }
 
+        if (!res._header) {
+          res._implicitHeader()
+        }
+
         if (chunk == null) {
           ret = true;
           return ret;
--- a/test/session.js
+++ b/test/session.js
@@ -1449,6 +1449,42 @@ describe('session()', function(){
       .get('/')
       .expect(200, 'hello, world', done)
     })
+
+    it('should error when res.end is called twice', function (done) {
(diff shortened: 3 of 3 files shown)


josh-byster:
After dropping this in-place and rerunning my existing code that caused me to discover this, it seems to be working exactly as I would expect! Thank you very much for this prompt fix!

dougwilson:
No problem at all, thanks for reporting!

PR #771, opened 2020-07-16
Browsers won't set a cookie in domains other than the server domain, Because it will be seen as an invalid domain.

prashanshan6:
> A cookie for a domain that does not include the server that set it should be rejected by the user agent.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie


dougwilson:
Thank you, but there is a limit to how much we are going to realistically document these type of things. This is why the documentation is currently set to "it sets the attribute to the value you provided". The main reason is that different web browsers end up implementing slightly different rules and those rules change over time.

PR #777, opened 2020-08-26
Add note about Aug 2020 Chrome updates that impact deliverability of …

john-redd:
…cross-site cookies.


dougwilson:
Thank you, but we don't really use this README to make web browser vendor announcements. If this is a note because it is specified in a web standard, we can certainly add a note to that, but web browsers are likely to have various differences with each other, and keeping up with them all is not really a great use of time for each and every npm module that deals with cookies. A more central place is likely better, perhaps the documentation the web browser vendor makes available.

PR #778, opened 2020-08-26
Note for new cookie spec

john-redd:


README.md
```diff
@@ -110,6 +110,10 @@ More information about the different enforcement levels can be found in
 **Note** This is an attribute that has not yet been fully standardized, and may change in
 the future. This also means many clients may ignore this attribute until they understand it.
 
+**Note** As of August 2020, Chrome versions 80 and above have implemented a new cookie [spec](https://tools.ietf.org/html/draft-west-cookie-incrementalism-00#page-4)
```
dougwilson, line 113: As noted in the previous PR, please remove reference to any specific web browser, as this is just going to lead to confusion, and much to keep up with. For example, this statement is not even correct, as Chrome 80 is not when the setting was introduced, nor turned on by default for users -- Google controls those outside of specific browser release versions. It of course doesn't explain Edge, Firefox, Safari, etc. and the other multitude of browsers all likely to make (or have made) similar changes.

john-redd: Removed the Chrome specific reference, leaving only a general reference to the spec. Sorry about the confusion with Chrome specific wording. Just had a hard time the last day trying to figure out why a project I was working on was no longer allowing the session cookie to be set client side in Chrome, but was allowing it to be set in Firefox.

dougwilson: Gotcha. I also noticed going to that draft, that it has been obsoleted and replaced with an updated one https://tools.ietf.org/html/draft-west-cookie-incrementalism-01 . Looking at the diffs, the updated version seems to be Chrome's current behavior, so probably want to link to the updated one instead.

john-redd: Thanks for catching that, just updated to the link to use the current draft link.

dougwilson (approved):

PR #813, opened 2021-02-01
Changed keywords to ES6 conventions

Deveshb15:
Changed keywords to ES6 conventions for better convenience in README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ $ npm install express-session
 ## API
 
 ```js
-var session = require('express-session')
+const session = require('express-session')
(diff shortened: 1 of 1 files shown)


marsonya (commented):
Express (and sister projects) wishes to support all node versions going back to 0.10.
In order to do so, it cannot support ES6 since ES6 syntax doesn't work on older Node versions.

To ensure the examples/snippets run for all node versions that express (and sister projects) support, ES6 cannot be used.
That's the reason you won't find any ES6 syntax anywhere.

Issue #725, opened 2020-02-02
using secure=auto with sameSite=none


nhitchins:
With the upcoming chrome 80 release and the need to set sameSite=none for cross-browser requests it would be good to consider the scenario where secure is set to 'auto' (set based on 'x-forwarded-proto' header). 
Chrome will start rejecting non 'secure' SameSite=None cookies. Therefore if the 'secure' attribute is not being set based on incoming request then sameSite=None option should revert to 'Lax'.

dougwilson:
Hm. I wonder if (1) is this only Chrome that does this, or it is all web browser (perhaps in a RFC spec somewhere) and (2) I don't think of course that if the user put `sameSite: 'none'` it should suddenly become lax; there would need to probably be some other kind of value to use instead to indicate you just want like "as high as possible" or something.

dougwilson:
Taking a look to remind myself what SameSite=none does, I think that it's not possible to do what you're asking. This is because SameSite=none means you need you cookie available cross-site, but the other values (like Lax) do not allow that. So it would seem that if this module just changed the value, your site's functionality would fundamentally be broken, no?

What would be the use-case where you would want it to be SameSite=none when secure but something different when unsecure? How would the cross-site behavior work in the unsecure case?

nhitchins:
@dougwilson 
I think chrome is the first to enforce it and other browsers will follow. Check this link for details:
https://www.chromium.org/updates/same-site
https://www.chromestatus.com/feature/5633521622188032

The scenario where this is required: 
A production site that allows itself to be embedded as an iframe at any origin but still uses cookies for authentication (sameSite=None required in Chrome 80). The production server(s) are behind an SSL proxy and we use proxy=true and secure='auto' options so that setting 'secure' is determined per-request by the 'x-forwarded-proto' header if it exists. This allows public access via the proxy (sameSite=None; secure) with full functionality. And it allows indirect access behind the SSL proxy (think internal status pages, etc.) with (sameSite=Lax) but these pages cannot be embedded in an iframe on a different origin which is fine.

nhitchins:
suggest a validation check to ensure if 'secure' is not then do not allow 'sameSite=None' as this combo will be rejected in chrome 80+. Instead of triggering an exception, just fallback to 'sameSite=Lax' ?

dougwilson:
[...]

PR #817, opened 2021-02-26
fix: secure='auto' and sameSite='none' case

simllll:
if secure is set to `auto` and sameSite is set to `none` it can result into secure is set to false, and therefore the cookie will not be set.
Because sameSite none always needs a secure context.

This fixes this behaviour, in case secure evaluates to false and sameSite is set explicitly to none, sameSite will be removed.

related to #778

### DIFF
--- a/index.js
+++ b/index.js
@@ -162,6 +162,11 @@ function session(options) {
 
     if (cookieOptions.secure === 'auto') {
       req.session.cookie.secure = issecure(req, trustProxy);
+      // if sameSite is set to "none", secure is required,
+      // but auto can set secure to false, therefor remove sameSite attribute in this case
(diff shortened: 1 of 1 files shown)


dougwilson:
See discussion in #725

simllll:
Sorry i missed that, still don't really follow the reasoning because secure has a auto flag, why samesite shouldn't benefit from that too... ;) But if I find time i will add a comment on the other thread. Thanks for pointing me to it :)!

dougwilson:
The issue is these two settings control completely different behaviors. Beaides that, if you need a samesite none cookie, how can your aite possibly work when the cookie is now lax or strict? It would reason that if you site does indees work with a strict or lax cookie, that is what you should be using and not none.

PR #833, opened 2021-07-07
Removes redundant check in session shouldTouch

nickshoe:
The shouldSave check is already performed in the caller (line 334) in an "either ... or ..." fashion.

### DIFF
--- a/index.js
+++ b/index.js
@@ -457,7 +457,7 @@ function session(options) {
         return false;
       }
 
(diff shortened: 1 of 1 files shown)


dougwilson:
Hello, and thank you for your pull request. I understand what you're saying, but these are independent functions. Leaving this function with an incomplete implementation will just be confusing as it would be unusable without knowing the other has to be called first.

PR #595, opened 2018-06-18
implement max duration of session

justinwatkinsact:
#557 code changes, unit tests and readme.md updates to add this feature.

PR #835, opened 2021-07-25
Implement `checkTouchModified` option, improve hashing performance, upgrade eslint-plugin-markdown

shahyar:
I've taken the idea from https://github.com/expressjs/session/pull/531 and done it in a way which I feel resolves the concerns from that thread.

Commit 1 modifies the hashing mechanism so that it doesn't get called multiple times per request, by caching the result. Commit 2 updates `eslint-plugin-markdown` to solve a security issue.

Commit 3 contains the new `checkTouchModified` option. If set to `true`, `session.touch()` does not get automatically called on every request. Instead, `shouldTouch` now checks if the session content or expiry has changed before calling `store.touch(...)`. This allows one to manually call `session.touch()` elsewhere in the application.

### DIFF
--- a/README.md
+++ b/README.md
@@ -269,6 +269,19 @@ will add an empty Passport object to the session for use after a user is
 authenticated, which will be treated as a modification to the session, causing
 it to be saved. *This has been fixed in PassportJS 0.3.0*
 
+##### checkTouchModified
+
+If set to `false`, unmodified sessions are always touched. However, when
+`rolling` is disabled, the expiration date in the cookie can differ from that
+which is in the store, since the cookie expiration date is only updated if the
+session is modified, but the session in the store is updated regardless.
+
+If set to `true` and the store supports "touching", unmodified sessions are not
+touched in the store at the end of each request. Manually calling
+`session.touch()` will extend the cookie expiry, and call `store.touch()`, which
+should implement server-side expiry extension.
+
+The default value is `false`.
 ##### secret
 
 **Required option**
--- a/index.js
+++ b/index.js
@@ -70,6 +70,7 @@ var defer = typeof setImmediate === 'function'
  * Setup session store with the given `options`.
  *
  * @param {Object} [options]
+ * @param {Boolean} [options.checkTouchModified] Enable check preventing touch if session is unmodified
  * @param {Object} [options.cookie] Options for cookie
  * @param {Function} [options.genid]
  * @param {String} [options.name=connect.sid] Session ID cookie name
@@ -86,6 +87,9 @@ var defer = typeof setImmediate === 'function'
 
 function session(options) {
   var opts = options || {}
+
+  // get the touch modification checker option
+  var checkTouchModified = opts.checkTouchModified
 
   // get the cookie options
   var cookieOptions = opts.cookie || {}
@@ -207,6 +211,7 @@ function session(options) {
 
     var originalHash;
     var originalId;
+    var originalExpires;
     var savedHash;
     var touched = false
 
@@ -228,8 +233,10 @@ function session(options) {
         return false;
       }
 
+      // Instead of getting the hash multiple times, only call when needed, and then cache it
+      var hashCache;
       function getHash() {
-        return hash(req.session);
+        return hashCache ? hashCache : (hashCache = hash(req.session));
       }
 
       if (!shouldSetCookie(req, getHash)) {
@@ -242,7 +249,7 @@ function session(options) {
         return;
       }
 
-      if (!touched) {
+      if (!touched && !checkTouchModified) {
         // touch session
         req.session.touch()
         touched = true
@@ -340,7 +347,7 @@ function session(options) {
         return _end.call(res, chunk, encoding);
       }
 
-      if (!touched) {
+      if (!touched && !checkTouchModified) {
         // touch session
         req.session.touch()
         touched = true
@@ -349,7 +356,7 @@ function session(options) {
       // Instead of getting the hash multiple times, only call when needed, and then cache it
       var hashCache;
       function getHash() {
-        return hashCache ? hashCache : hash(req.session);
+        return hashCache ? hashCache : (hashCache = hash(req.session));
       }
 
       if (shouldSave(req, getHash)) {
@@ -385,6 +392,7 @@ function session(options) {
       store.generate(req);
       originalId = req.sessionID;
       originalHash = hash(req.session);
+      originalExpires = req.session.cookie.expires;
       wrapmethods(req.session);
     }
 
@@ -393,6 +401,7 @@ function session(options) {
       store.createSession(req, sess)
       originalId = req.sessionID
       originalHash = hash(sess)
+      originalExpires = req.session.cookie.expires
 
       if (!resaveSession) {
         savedHash = originalHash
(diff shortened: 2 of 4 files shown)


dougwilson (changes requested):
I know that thread ia long, but history of thia module has demonstrated that adding options that allow the user to alter how the API works between this module and the store modules results in strange behaviors with the stores. The touch is working as intended currently, being called if save is not (so between save and touch there will always be a call in to the store).

I think the best thing is let's just add a new API to the stores to achieve what you are trying to do and leave save and touch working as they are today. Then the stores will have these different signals available to achieve whatever it is that you are trying to achieve.

dougwilson:
> checks if the session [...] expiry has changed

This module has no knowledge of session expiry, only the expiration of the cookie. The session expiry is managed only by the store implementations. This is actually the main reason touch API was added so that there will _always_ be a call into the store even if nothing was changed in the session so the stores have the opportunity to update expiry if they want.

shahyar:
> The touch is working as intended currently, being called if save is not

Right, that's fair, but I have to argue that while that may be intended, it's not working optimally nor securely. It's not at all clear that `touch` is getting triggered on every request. That is only documented in the code. At a minimum, `touch` should be told if the session has changed (it currently receives no arguments). As was mentioned in the other thread, the idea of `rolling` makes the matter even more confusing.

As an example, with default settings, one might think they have set a maxAge of 30 days at login. No further writes should occur. But the session is in fact being written every time due to this call. Yes, you could simply make your store's `touch` a noop, but then there is no way to manually call `session.touch()`, because `session.touch` does not pass its arguments to `store.touch`, to do something like `touch(force = false)` in a custom implementation.

> This module has no knowledge of session expiry 

Sure it does. It checks on line 472 that `req.session.cookie.expires` is not null.  This only only cares if that value changed, and thus considers the session dirty.

dougwilson:
> Sure it does. It checks on line 472 that req.session.cookie.expires is not null. This only only cares if that value changed, and thus considers the session dirty.

Like I said, it checks the cookie expiry, not the sessiom expiry. Sessiom expiry (and thus max sessiom length you mentiomed above), is handled by the store you are using. A user can just save the value of a cookie and send it any time, ignoring a cookie expiry. You cannot trust the cookie expiry to end your sessions, you meed to store a date in your database and limit it there.

dougwilson:
I woupd help come up with a solution, but it is not clear what you are trying to achieve. The way touch works is nkt going to change, as it will break our API contract with existing store modules.

If you can explain what you are trying to do with your store module, I'm sure we can come up with whatever additional store API is needed to convery the information to the store in order to achieve the desired outcome.

dougwilson:
> At a minimum, touch should be told if the session has changed (it currently receives no arguments).

It doesn't need to, as touch() is only called is a session was not modified, and save() is called if the session is modified. Whcih one of those is called is the signal for is the session was modified or not. It only calls one or the other for a session to the store.

shahyar:
> touch() is only called is a session was not modified

Yeah, that was a poor suggestion on my part.

> Like I said, it checks the cookie expiry, not the sessiom expiry 

Sorry, perhaps the way I worded it was misleading. I do not check the session's expiry (as it is not available here). I only check `req.session.cookie.expires` to see if it changed, as a shortcut to not having to serialize the entire session object to compare if _that_ changed.

> I woupd help come up with a solution, but it is not clear what you are trying to achieve. The way touch works is nkt going to change, as it will break our API contract with existing store modules.

I've added tests to ensure it does not break the store API. Enabling this option would simply leave calling `touch` to the implementation, rather than being called every time.

To clarify: I want to be able to manually extend (touch) my sessions. To do this without modifying express-session, I would have a couple of options:

1. Don't implement `touch` (or define it as noop). Then, directly modify the store to extend the expiry when I need to.
2. Hijack the `Session.prototype` to either modify `touch` or add my own method onto it, which I can call.

Both of these are not great. The alternative solution is to be able to pass my own `options.session=Session` definition, like with `options.store=MemoryStore`.

Though with my current proposal, no existing codepaths will break. In fact, as can be seen on the other PR, the popular `connect-redis` as well as `connect-session-sequelize` have a `disableTouch` option to straight up disable `touch`. That should be indicative of a problem here. But neither offer a way to allow users to do it themselves, as `Session` does not provide any such extendibility.

dougwilson:
> I've added tests to ensure it does not break the store API.

To clarify: if a user can set an option that changes when save() and touch() are called to the store, that is breaking the store API, no? The touch() should be called to the store when a session was not modified and the save() should be called when it was modified. One or the other should always be called to the store so the store has the opportunity to perform whatever logic it needs based on how it's database works.

> have a disableTouch option to straight up disable touch. That should be indicative of a problem here.

I mean, at best it is an indication of a problem with that store implementation or just a poor naming of an option in the store. If users don't want the store to do whatever it ia doing to unmodified session, why is the store doing it?

shahyar:
So, then as we can't come to an agreement on this topic, I'd jump back to the alternate proposal: that `Session` can be supplied as an option, so that `req.session.doWhateverIWant` could be implemented. What do you think about that?

dougwilson:
We could certainly add that, but I'm not sure that would work with the store or not. But it's hard to really say what is a good solution or not without really understanding what you are trying to achieve, exactly.

dougwilson:
As for your example, can't you already define whatever you want like that with `require('express-session').Session.prototype.doWhateverIWant = function () {}` ?

shahyar:
> As for your example, can't you already define whatever you want like that with require('express-session').Session.prototype.doWhateverIWant = function () {} ?

As I said in my last comment, hijacking the prototype is pretty bad practice.

> without really understanding what you are trying to achieve, exactly

The two stores mentioned above as well as the one I'm using all update the database on `touch`. I'm trying to avoid that.

dougwilson:
> As I said in my last comment, hijacking the prototype is pretty bad practice.

That sounds pretty subjective to me. I don't see why that is bad at all. Can you link to what makes that bad, exactly? This module exposes the reference for usage and extension. Maybe explicitly documenting it like was done with Express would make it more "official" or somethin?

> The two stores mentioned above as well as the one I'm using all update the database on touch. I'm trying to avoid that. But I still want to use req.session to update the expiry, without updating the content.

I get that, but of course you probably realize you are describing a store issue, not an issue with this module. This module signals to the store if the session was modified and should be saved (with save() call) or not modified (with touch() call). Rather than adding more complication here trying to prevent this module from telling the store this session was not modified (by skipping touch() call to store), it seems much more realistic to change what the undesired behavior is in the store. You could even use the decorator pattern to pass in a decorated store that does whatever you want with unmodified sessions. AOP is another programming method that could be used as well. But it sounds like you're trying to "trick" the stores by preventing the unmodified notification because you don't like what they are doing with that notification.

shahyar:
I don't know what else to say. I've explained my reasoning quite well.

It seems you aren't particularly open to any feedback on this project. https://github.com/expressjs/session/pull/595 has been open for years. Comments were left on the previous PR by multiple people. That issue was referenced in several other projects. It seems we are all wrong.

So, I'll just close this and keep my fork. Thanks.

dougwilson:
> It seems you aren't particularly open to any feedback on this project. 

You're certainly entitled to your opinion, but hey, I guess maintaining and giving this project away for free is just not enough these days, even with a license like MIT.

> #595 has been open for years. Comments were left on the previous PR by multiple people.

Sure, and the author of the PR never made the changes requested, so it sat there waiting for the author to update the PR. That is on the PR author, not this project.

PR #839, opened 2021-08-05
Corrected typographical errors

olayeni45:
Corrected the typographical errors for the saveUninitialized documentation and the resave documentation in the README file.


dougwilson:
The original word was correct and not a typo. https://en.m.wikipedia.org/wiki/Race_condition

olayeni45:
Oh sorry about that, didn't know. Thanks for the information.

PR #844, opened 2021-09-06
test: make tests pass on Node.js >=16.9.0

targos:
With V8 9.3, the error message is: "Cannot read properties of undefined (reading 'expires')"

See https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker/2754/nodes=fedora-latest-x64/testReport/junit/(root)/citgm/express_session_v1_17_2/


dougwilson (approved):
Thank you for this. Sorry for the delay in merging.

PR #849, opened 2021-10-18
fix: skip re-saving session when not modified since save

jneander:
When the `saveUninitialized` option is set to false, sessions will be
saved at the end of the request even when they have already been saved
earlier in the request and not subsequently modified. The result is a
double save that can potentially be wasteful with the persistence layer.
This change checks for modification of the session before saving, and
saves only if needed.

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,6 +1,7 @@
 unreleased
 ==========
 
+  * Fix resaving already-saved new session at end of request
   * deps: cookie@0.4.2
 
 1.17.2 / 2021-05-19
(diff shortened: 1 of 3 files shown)


jneander:
I'm boggled by the Node.js 0.8 failure at https://github.com/expressjs/session/pull/849/checks?check_run_id=3929683024. Any ideas?

jneander:
Thank you, Doug. I appreciate that.

dougwilson (approved):
Sorry for the long delay :(

Issue #857, opened 2021-12-22
Enable secure cookies over localhost


thegoldgoat:
I think it is reasonable to enable secure cookies on localhost, since browsers accepts them.

This simplifies development and avoid developers confusion, since now cookies are suppressed silently.

What I propose is updating isSecure:

https://github.com/expressjs/session/blob/0048bcac451ad867299d404aca94c79cc8bc751d/index.js#L623

And add something like this:

```js
if (req.connection.remoteAddress === '127.0.0.1' ||
    req.connection.remoteAddress === '::ffff:127.0.0.1'
) {
    return true;
}
```

I am opening a PR in a moment

Also see #837

dougwilson:
Hello, and thanks for opening this issue. It seems to be a duplicate of #837 though. I can re-open if there is elaboration on how it is not a duplicate, though.

PR #858, opened 2021-12-22
Add secure cookies when remote is localhost

thegoldgoat:
See #857

This PR enables secure cookies to be transmitted when the remoteAddress is localhost (both IPv4 & IPv6)

I also updated the README to match the new behavior, but I need help writing the tests: the one marked 'TODO' currently fails, we would need to create a couple of tests for both localhost connection and non-localhost one.

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,10 +125,10 @@ the cookie back to the server in the future if the browser does not have an HTTP
 connection.
 
 Please note that `secure: true` is a **recommended** option. However, it requires
-an https-enabled website, i.e., HTTPS is necessary for secure cookies. If `secure`
-is set, and you access your site over HTTP, the cookie will not be set. If you
-have your node.js behind a proxy and are using `secure: true`, you need to set
-"trust proxy" in express:
+an https-enabled website (or connection via localhost), i.e., HTTPS is necessary
+for secure cookies. If `secure` is set, and you access your site over HTTP, the 
+cookie will not be set. If you have your node.js behind a proxy and are using
+`secure: true`, you need to set "trust proxy" in express:
 
 ```js
 var app = express()
--- a/index.js
+++ b/index.js
@@ -621,6 +621,14 @@ function hash(sess) {
  */
 
 function issecure(req, trustProxy) {
+
+  // socket is localhost
+  if (req.connection.remoteAddress === '127.0.0.1' ||
+      req.connection.remoteAddress === '::ffff:127.0.0.1'
+  ) {
+    return true;
+  }
+
   // socket is https server
   if (req.connection && req.connection.encrypted) {
     return true;
--- a/test/session.js
+++ b/test/session.js
@@ -610,6 +610,7 @@ describe('session()', function(){
         .expect(200, done)
       })
 
+      // TODO: Fix tests
       it('should work when no header', function(done){
         request(server)
         .get('/')


thegoldgoat:
**Still needs to update tests**, marked as draft.

index.js
```diff
@@ -621,6 +621,14 @@ function hash(sess) {
  */
 
 function issecure(req, trustProxy) {
+
+  // socket is localhost
+  if (req.connection.remoteAddress === '127.0.0.1' ||
+      req.connection.remoteAddress === '::ffff:127.0.0.1'
+  ) {
```
rhansen, line 628: Won't this condition be true for most reverse proxy setups? Or does Express mutate `req.connection.remoteAddress` if `trust proxy` is enabled? (The documentation is not clear.)

index.js
```diff
@@ -621,6 +621,14 @@ function hash(sess) {
  */
 
 function issecure(req, trustProxy) {
+
+  // socket is localhost
+  if (req.connection.remoteAddress === '127.0.0.1' ||
+      req.connection.remoteAddress === '::ffff:127.0.0.1'
```
rhansen, line 627: What about IPv6 localhost?

```suggestion
      req.connection.remoteAddress === '::ffff:127.0.0.1' ||
      req.connection.remoteAddress === '::1'
```

index.js
dougwilson: You are correct. I don't think this will work correctly as any reverse proxy on localhost (think typicall ssl ofloading or even just kubernetes side cars) would end up as thinking this is a localhost connection. Express does not alter any values that are provided from Node.js HTTP core.

README.md
```diff
@@ -125,10 +125,10 @@ the cookie back to the server in the future if the browser does not have an HTTP
 connection.
 
 Please note that `secure: true` is a **recommended** option. However, it requires
-an https-enabled website, i.e., HTTPS is necessary for secure cookies. If `secure`
-is set, and you access your site over HTTP, the cookie will not be set. If you
-have your node.js behind a proxy and are using `secure: true`, you need to set
-"trust proxy" in express:
+an https-enabled website (or connection via localhost), i.e., HTTPS is necessary
+for secure cookies. If `secure` is set, and you access your site over HTTP, the
```
dougwilson, line 129: > HTTPS is necessary
for secure cookies

Probably this line would need reworded with a change like this, as I believe that is the whole point of this change, that HTTP would work for localhost.

index.js
thegoldgoat: Oh, I forgot about that you are right!

index.js
thegoldgoat: Yeah I see, I think the reasonable way to do this is to check  if we are behind a proxy

index.js
```diff
@@ -624,9 +624,20 @@ function issecure(req, trustProxy) {
...
+      req.connection.remoteAddress === '::1'
   ) {
-    return true;
+    // if proxy is trusted; localhost connection is secure for sure
+    if (trustProxy === true) {
+      return true;
+    }
+
+    // proxy not explicitly trusted; no proxy means connection is secure
+    if (req.headers['x-forwarded-proto'] !== undefined) {
```
dougwilson, line 636: Unfortunately this is also not a reliable test, because if the user does not trust the connection as a proxy, this header may not be set if not a proxy or may be forgable by the client.

README.md
thegoldgoat: Take a look at 8314c37, I think it is clear that way

dougwilson: Thanks. I left a comment already on it 👍

README.md
```diff
@@ -124,11 +124,16 @@ attribute is not set.
...
-an https-enabled website, i.e., HTTPS is necessary for secure cookies. If `secure`
-is set, and you access your site over HTTP, the cookie will not be set. If you
-have your node.js behind a proxy and are using `secure: true`, you need to set
-"trust proxy" in express:
+Please note that `secure: true` is a **recommended** option. However, for the cookie
+to be set, it requires one of these conditions:
+
+  - https-enabled website
+  - localhost connection
+  - node.js behind a proxy and "trust proxy" in express
```
dougwilson, line 132: It wouldn't just be behind trust proxy, as you can still have a http connection though a proxy.

index.js
thegoldgoat: Sorry I am not sure what you mean, maybe I am missing some possible network configuration?

Since we detect a localhost connection we could trust the remote (reverse proxy or client) right?

Please correct me if I am wrong

dougwilson: Not unless the user of this module specifies they trust it. This fixed a reported security vulnerability.

thegoldgoat: Oh I see what you mean, so we can conclude that the only thing we can do is return true if the proxy is trusted and the connection is localhost, right?

README.md
thegoldgoat: Yeah my bad, does `node.js behind an HTTPS proxy and "trust proxy" in express` sound good?
I also need to update the localhost one, specifying that "trust proxy" is also needed

thegoldgoat:
This has been stale for a while...

 > Oh I see what you mean, so we can conclude that the only thing we can do is return true if the proxy is trusted and the connection is localhost, right?

So, are you ok with this one? Otherwise I would have no other idea and we can close this..

PR #859, opened 2021-12-24
build: Upper-case `GITHUB_TOKEN`

rhansen:
[The documentation][1] says, "Secret names are not case-sensitive." However, using upper-case letters improves readability by avoiding questions about whether lower-case is acceptable. It also matches the capitalization used earlier in the file (line 168) and in documentation.

[1]: https://docs.github.com/en/actions/security-guides/encrypted-secrets#naming-your-secrets

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -176,5 +176,5 @@ jobs:
     - name: Upload code coverage
       uses: coverallsapp/github-action@master
       with:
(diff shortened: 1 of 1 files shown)


dougwilson:
Thank you, and I guess if you are saying they are equivalent then maybe they are, but ultimately we are following the official docs from coveralls: https://github.com/coverallsapp/github-action#complete-parallel-job-example

PR #860, opened 2021-12-24
build: Drop support for unmaintained Node.js versions

rhansen:
This makes it possible to use modern ECMAScript features.


dougwilson:
The modules for Express are designed to support the same versions the current main line express supports.

dougwilson:
If there is a feature that requires a newer version of Node.js that would change the external API of this module, that can certainly be considered to diverge from that, as it would provide a clear benefit to the usage of this module.

PR #861, opened 2021-12-24
Cookie: Fix originalMaxAge option corruption

rhansen:
This is necessary to survive a round trip to JSON. (But it is not sufficient—the `expires` option must be converted from a string to a `Date` object. This will be addressed in a future pull request.)

### DIFF
--- a/session/cookie.js
+++ b/session/cookie.js
@@ -32,7 +32,26 @@ var Cookie = module.exports = function Cookie(options) {
       throw new TypeError('argument options must be a object')
     }
 
+    var keys = [];
     for (var key in options) {
+      keys.push(key);
+    }
+    keys.sort(function(a, b) {
+      var relativeOrder = {
+        // The setter for .maxAge modifies .expires, so .maxAge must be set before .expires if both
+        // are present in options.
+        maxAge: 0,
+        // The setter for .expires modifies ._expires and .originalMaxAge, so it must be set before
+        // those.
+        expires: 1,
+        _expires: 2,
+        originalMaxAge: 2,
+      };
+      if (!(a in relativeOrder) || !(b in relativeOrder)) return 0;
+      return relativeOrder[a] - relativeOrder[b];
+    });
+    for (var i = 0; i < keys.length; i++) {
+      var key = keys[i];
       if (key !== 'data') {
         this[key] = options[key]
       }
--- a/test/cookie.js
+++ b/test/cookie.js
@@ -56,6 +56,19 @@ describe('new Cookie()', function () {
         assert.strictEqual(cookie.expires, expires)
       })
 
+      it('should not override originalMaxAge option', function () {
+        // This test relies on for-in iteration order, but for-in iteration order is only specified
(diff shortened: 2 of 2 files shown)


dougwilson:
Hello, and thank you for your pull request. I'm not quite clear what you are demonstrating here, can you elaborate more? The `Cookie` constructor is not a public API (see https://github.com/expressjs/session for the public API) and is simply for the internal use of this module. Is this fixing a bug in some store module perhaps? Maybe is it possible to demonstrate what the bug is when only using the public API + some store module to better elaborate?

dougwilson:
As for your comment about expires being a string, you are correct, but the string to `Date` conversion happens here: https://github.com/expressjs/session/blob/66634e9f77132a444dfda7e279a399054e3b28df/session/store.js#L92

dougwilson:
And the proper ordering of `originalMaxAge` when loading out of the store happens here: https://github.com/expressjs/session/blob/66634e9f77132a444dfda7e279a399054e3b28df/session/store.js#L97-L98

dougwilson:
It also looks like this change will be backwards-incompatible as well, as it alters the documented working of the `cookie` option in the public API. https://github.com/expressjs/session#cookiemaxage

> Note If both expires and maxAge are set in the options, then the last one defined in the object is what is used.

I suggest perhaps opening an issue describing the issue at hand so we can better understand it. We can always work together on a possible fix ideas and open a PR from there. It seems like a PR is premature at this point, as the issue is not clear and the proposal here breaks public API.

rhansen:
Sorry for the noise!

I was troubleshooting a problem we had in our custom session store, and thought I had figured it out. I thought it was because the session object wasn't surviving a round-trip through JSON due to two things: (1) the Cookie constructor wasn't converting the `expires` string to a `Date` object, and (2) the property order of `Cookie.prototype.toJSON()` and the use of `for...in` iteration effectively caused the `originalMaxAge` property to be ignored.

But it turns out my diagnosis was wrong. I tried writing up a minimal reproducing example, but couldn't thanks to the code you pointed out. So the problem we're experiencing is somewhere else, probably in our code.

From a code readability perspective, I think the conversion from string to `Date` and the `originalMaxAge` order issue should be solved in the Cookie constructor (or in a static `Cookie.fromJSON()` function), not in `Store.prototype.createSession()`. But the current code behaves correctly as far as I can tell.

> It also looks like this change will be backwards-incompatible as well, as it alters the documented working of the `cookie` option in the public API. https://github.com/expressjs/session#cookiemaxage
> 
> > Note If both expires and maxAge are set in the options, then the last one defined in the object is what is used.

Just an FYI, `for...in` iteration was not guaranteed to iterate in property definition order until ES2020. (I don't know how historic Node.js versions behaved, so maybe this is pedantic.)

dougwilson:
> Just an FYI, for...in iteration was not guaranteed to iterate in property definition order until ES2020. (I don't know how historic Node.js versions behaved, so maybe this is pedantic.)

V8 (and thus all versions of Node.js) guarantees the order. ECMAScript just eventually defined V8's behavior as official in 2020 version.

PR #869, opened 2022-01-14
Removing unnecessary files in package.json

noahliechti:
To state the npm documentation:

```
Certain files are always included, regardless of settings:

    package.json
    README
    CHANGES / CHANGELOG / HISTORY
    LICENSE / LICENCE
    NOTICE
    The file in the "main" field

README, CHANGES, LICENSE & NOTICE can have any case and extension.
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,7 +32,6 @@
   "files": [
     "session/",
     "HISTORY.md",
-    "LICENSE",
     "index.js"
   ],
   "engines": {


dougwilson:
Thank you. These two files are kept in the package on purpose.

dougwilson:
Sorry, I misunderstood what this PR was doing.

import-brain (approved):

package.json
```diff
@@ -31,8 +31,6 @@
   },
   "files": [
     "session/",
-    "HISTORY.md",
```
dougwilson, line 34: `npm pack` is not including this file after this change:
```
$ npm pack
npm notice
npm notice 📦  express-session@1.17.2
npm notice === Tarball Contents ===
npm notice 1.2kB  LICENSE
npm notice 40.5kB README.md
npm notice 16.1kB index.js
npm notice 1.3kB  package.json
npm notice 2.6kB  session/cookie.js
npm notice 3.6kB  session/memory.js
npm notice 3.0kB  session/session.js
npm notice 1.9kB  session/store.js
npm notice === Tarball Details ===
npm notice name:          express-session
npm notice version:       1.17.2
npm notice filename:      express-session-1.17.2.tgz
npm notice package size:  18.4 kB
npm notice unpacked size: 70.2 kB
npm notice shasum:        d54407aac54315fc965a34fa47c95a291773e192
npm notice integrity:     sha512-LCVOIaWg8QEZE[...]OnYyYKLJ8ZOeg==
npm notice total files:   8
npm notice
express-session-1.17.2.tgz
```

Could be a bug in npm or something. Can you put this one back at least for now, so it will be included in our pack?

dougwilson (changes requested):

package.json
dougwilson: Actually the current npm docs (https://docs.npmjs.com/cli/v8/configuring-npm/package-json#files) no longer lists `HISTORY` like you quote does. So it seems maybe it was removed by npm, so we'll need to keep it in the list here.

noahliechti: I'll add it again.

noahliechti: @dougwilson done

Issue #872, opened 2022-02-02
Feature request: configurable header name for x-forwarded-proto


rickknowles-cognitant:
Hi,

Would it be possible to make the exact header-name checked for in line 640 of index.js to be a config option instead of hard coded to "x-forwarded-proto" please ?

![image](https://user-images.githubusercontent.com/37247203/152174193-de7c0fe1-3337-435a-a9fd-f164fcb8f156.png)

The reason I ask is that we have a deployment situation where there are 2 AWS load balancers sequentially reverse proxying before the express code is requested, and the 2nd one unfortunately overwrites the x-forwarded-proto set by the 1st (which has the value we want). 

In other applications we've been able to write the 1st LB's x-forwarded-proto into another header so it's visible downstream, and then look at the other header. Unfortunately without forking and modifying express-session's source, we can't find a way to trick express-session's issecure() method into looking at that other named header, so all secure cookies are dropped.

I'm happy to fork and maintain a clone if its the only way, but I just thought I'd ask: if it's possible to take the header name to use instead of x-forwarded-proto as an option in the session constructor that could be overridden or a custom ```getXForwardedProtoHeaderValue()``` function be supplied it would make that unnecessary.

Happy to write a patch if it would help and submit, just wanting to sound out whether it's an option or not. Thank you,

Rick Knowles

dougwilson:
Using that header in this lib is an old fallback. You should instead use whatever logic you like to set `req.secure`, which rhis module will honor.

rickknowles-cognitant:
Isn't req.secure not settable though ?

rickknowles-cognitant:
And isn't closing this a little premature ?

dougwilson:
Hi, yes you can co figure it though express to use any logic you like. When an issue has a straight forward answer, we typically close it until new information comes to light, and many people don't respond back. Closing the issue does not stop comments, nor our replies, as you can see :)

rickknowles-cognitant:
OK sure - I'll I can see is this though:

![image](https://user-images.githubusercontent.com/37247203/152177018-f9f23244-20a0-4d77-876e-e8f8adb8cdfb.png)

and when we tried doing ```req.secure = true``` it doesn't allow it. How exactly would one go about configuring it with any logic we like ? Obviously I prefer this approach too, I've only asked for this feature because the approach wasn't allowed

dougwilson:
http://expressjs.com/en/guide/overriding-express-api.html

```js
[...]

PR #879, opened 2022-03-04
Support "Forwarded" header too

bschaepper:
Since there is a standardized replacement for `X-Forwarded-Proto`, the new [`Forwarded` header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Forwarded), we should probably support this too.

This change adds a check for the new header, after all preexisting checks for `issecure`. This allows to set secure cookies when a `Forwarded: proto=https` header is present.

### DIFF
--- a/README.md
+++ b/README.md
@@ -200,12 +200,12 @@ each other. The simplest method is to simply set different `name`s per app.
 
 ##### proxy
 
-Trust the reverse proxy when setting secure cookies (via the "X-Forwarded-Proto"
+Trust the reverse proxy when setting secure cookies (via the "X-Forwarded-Proto" or "Forwarded"
 header).
 
 The default value is `undefined`.
 
-  - `true` The "X-Forwarded-Proto" header will be used.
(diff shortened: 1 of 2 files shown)


dougwilson:
Hello, and thank you for your pull request. The reading of x-forwarded-for is deprecated in this module and will be removed in next major version; instead `req.secure` is preferred, as it can be programmed to use any header name. Adding this directory is unfortunately tricky, as servers who are using x-forwarded-for at the gateway are unlikely to have filtered out the forwarded header to the backend.

Basically a duplicate request of #872

bschaepper:
Hi @dougwilson, oh I wasn't aware of that. Looks like a better solution yes, it felt a bit clumsy to add indeed.

dougwilson:
Haha, yea. Also there are some plans to make the proxy stuff more configurable and pluggable directly in Express. I'm glad to see the Frowarded header start to ger adoped, it's about time :) Express direcrly only supported the x- for all this time because there was practically no reason to otherwise, lol.

PR #887, opened 2022-04-22
add support for alternative forwarded header from proxy

### LINKED ISSUE #886: Support "Forwarded" header as an alternative to "X-Forwarded-Proto"
### problem
Unfortunately AWS API Gateway is using the less standard header [Forwarded](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Forwarded) when proxying requests.
This looks like this:
```
forwarded: by=1.2.3.4;for=5.6.7.8;host=my.example.com;proto=https
via: HTTP/1.1 AmazonAPIGateway
```

### solution
It would be great if we could support this header as a fallback when `X-Forwarded-Proto` is not present.

### solution alternative
`issecure` can be overwritten or extended with a function passed through options.

### security considerations
An external request might define the forward header to simulate as if it had been proxied.
Like the `X-Forwarded-Proto`, `Forwarded` is a forbidden request header and should not be set by browser and be deleted by proxies.

nmalyschkin:
fix #886

### DIFF
--- a/index.js
+++ b/index.js
@@ -637,13 +637,30 @@ function issecure(req, trustProxy) {
   }
 
   // read the proto from x-forwarded-proto header
-  var header = req.headers['x-forwarded-proto'] || '';
-  var index = header.indexOf(',');
-  var proto = index !== -1
-    ? header.substr(0, index).toLowerCase().trim()
-    : header.toLowerCase().trim()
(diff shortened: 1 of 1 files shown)


dougwilson:
Duplicate of #879

PR #892, opened 2022-05-14
Session Touch Reduction Option

### LINKED ISSUE #891: `session.touch` is called on every request when `resave: false`
# Description

[PR incoming with unit tests and implementation of option - default to no change from current behavior]

`session.touch` is called on every request when `resave: false` is set.

On in-memory stores such as [connect-redis](https://www.npmjs.com/package/connect-redis) this has little impact other than additional traffic.

However, on disk-backed and/or globally replicated stores the costs can be significant, such as [connect-dynamodb](https://www.npmjs.com/package/connect-dynamodb)

# Example Scenario with Costs - DynamoDB Store

[DynamoDB Provisioned Capacity Pricing](https://aws.amazon.com/dynamodb/pricing/provisioned/)

image

- RCU = Read Capacity Unit
- WCU = Write Capacity Unit
- 60,000 RPM site / 60 seconds / minute = 1000 RPS
- Write on Every Request
  - 1,000 RPS requires
    - 1,000 RCU x $0.00013 = $0.13 / hour = $1,138 / year
    - 1,000 WCU x $0.00065 = $0.65 / hour = $5,964 / year
    - Total = `$7,102` / year
- Write on Update  or when 75% of `originalMaxAge` Expired
  - Assuming 1% of requests update the session or reach 75% of `originalMaxAge`
  - 1,000 RPS requires
    - 1,000 RCU x $0.00013 = $0.13 / hour = $1,138 / year
    - 10 WCU x $0.00065 = $0.0065 / hour = $57 / year
    - Total = `$1,195`
    - Savings = `83%` vs Write on Every Request

# Best cases for cost reduction
- Relatively longer sessions (e.g. minutes, hours, days, months, but not seconds)
- Session values infrequently updated

huntharo:
Fixes #891

### DIFF
--- a/index.js
+++ b/index.js
@@ -75,6 +75,7 @@ var defer = typeof setImmediate === 'function'
  * @param {String} [options.name=connect.sid] Session ID cookie name
  * @param {Boolean} [options.proxy]
  * @param {Boolean} [options.resave] Resave unmodified sessions back to the store
+ * @param {Boolean} [options.touchAtMaxAgeRatio] `.touch()` session when `maxAge` has reached `.touchAtMaxAgeRatio` of the `originalMaxAge`
  * @param {Boolean} [options.rolling] Enable/disable rolling session expiration
  * @param {Boolean} [options.saveUninitialized] Save uninitialized sessions to the store
  * @param {String|Array} [options.secret] Secret for signing session ID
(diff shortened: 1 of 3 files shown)


dougwilson:
The touch action is meant for any session activity. If you need some other type of mechanism, please open an issue with what you are trying to achieve and we can work on some interface for it. Just based on the referenced issue, it seems there is confusion on what the purpose of touch is for.

huntharo:
@dougwilson - I legit opened an issue and linked it the description :)

`touch` only behaves as described now because there are no options to control its behavior.  That is the point of this PR and Issue.

dougwilson:
Right, but that is the purpose of touch -- to be called every time there is activity on the session. If the store needs to throttle to it's storage mechanism, that is a store-specific concern that should be implemented in the store module.

PR #900, opened 2022-06-26
feat: add support for http2

yunnysunny:


dougwilson:
Hello, ans thank you for your pull request. It looks like this adds an option to this module that, when set, modifies the global Node.js environment.

Unfortunately global modification is not something we would accept as a feature to this module. But the good news is that since you are modifying the global environment, there is no need to add it to this module: simply make the modification in your code or even publish it to npm as it's own module to use.

lamweili:
Would PR #908 be a better solution?
It replaces the undocumented `res._implicitHeader` with its implementation and as a by-product, also enables HTTP2 support.

PR #907, opened 2022-07-29
Pass new session as extra param to Session reload and regenerate callbacks

sgpinkus:
Why? The scope calling regenerate (or reload) does not necessarily want or need to know what a `req` is. Passing the new session object as a callback parameter allows the callback to regenerate or resave then make changes without knowing about `req`. Example:

```
  session.regenerate((err, newSession) => {
    if(err) return next(err);
    session.justRegenerated = true;  // Just an example.
  });
```

I specifically wanted this to `Session.regenerate` but added to `Session.reload` too for consistency. Might also want it in `Session.save` callback but that seems to require change in the behaviour of plugins so I gave it a pass.

### DIFF
--- a/README.md
+++ b/README.md
@@ -341,7 +341,7 @@ a new SID and `Session` instance will be initialized at `req.session`
 and the `callback` will be invoked.
 
 ```js
-req.session.regenerate(function(err) {
+req.session.regenerate(function(err, session) {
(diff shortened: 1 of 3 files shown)


sgpinkus:
@import-brain, @dougwilson Any chance merge?

PR #908, opened 2022-08-02
refactor: replaced _implicitHeader

### LINKED ISSUE #888: No need to call undocumented nodejs function ServerResponse._implicitHeader
express-session call such code to ensure http header has been sent before writing data.
But nodejs has include such code in the ServerResponse class  `write` function.
 
Another side we should consider is that `http2.Http2ServerResponse` does not have the property of `_implicitHeader`. When we start server via http2, the express-session will break.

lamweili:
Fixes #888

Replaced the usage of undocumented HTTP API with its implementation instead (bringing about **_some_*** http2 support).
There is no change to the existing behavior. This is similar to https://github.com/expressjs/compression/pull/170.

- Undocumented `res._implicitHeader` https://github.com/nodejs/node/blob/a8bc202d855cc7d14895db38a2ac09d1f873e803/lib/_http_server.js#L281
Replaced with `res._implicitHeader` implementation
https://github.com/nodejs/node/blob/a8bc202d855cc7d14895db38a2ac09d1f873e803/lib/_http_server.js#L282

- Undocumented `res._header` 
https://github.com/nodejs/node/blob/93e0bf9abf350637d772bcce14a5d9527b733300/lib/_http_outgoing.js#L741
Replaced with `res.headersSent`
https://github.com/nodejs/node/blob/93e0bf9abf350637d772bcce14a5d9527b733300/lib/_http_outgoing.js#L745

The changes do not require a semver major. 

_**some**_* - This change may/may not provide full http2 support.

### DIFF
--- a/index.js
+++ b/index.js
@@ -272,7 +272,7 @@ function session(options) {
           return ret;
         }
 
-        if (!res._header) {
+        if (typeof res.headersSent === 'boolean' ? !res.headersSent : !res._header) {
           res.writeHead(res.statusCode)
         }
 
--- a/index.js
+++ b/index.js
@@ -273,7 +273,7 @@ function session(options) {
         }
 
         if (!res._header) {
-          res._implicitHeader()
+          res.writeHead(res.statusCode)
         }
 
         if (chunk == null) {
--- a/test/session.js
+++ b/test/session.js
@@ -1381,6 +1381,25 @@ describe('session()', function(){
     })
   });
 
+  describe('res._header patch', function () {
+    it('should be equivalent to res.headersSent', function (done) {
+      request(createServer(function(req, res) {
+        assert.strictEqual(!!res._header, !!res.headersSent,
+          'res._header should be equal to res.headersSent (prior state change)')
+
+        var oldState = !!res._header;
+        res.end('ended')
+        var newState = !!res._header;
+        assert.notStrictEqual(oldState, newState)
+
+        assert.strictEqual(!!res._header, !!res.headersSent,
+          'res._header should be equal to res.headersSent (after state change)')
+      }))
+      .get('/')
+      .expect(200, 'ended', done)
+    })
+  })
+
   describe('res.end patch', function () {
     it('should correctly handle res.end/res.write patched prior', function (done) {
       function setup (req, res) {
--- a/test/session.js
+++ b/test/session.js
@@ -2313,10 +2313,7 @@ describe('session()', function(){
       })
       server.on('listening', function () {
         var client = createHttp2Client(server.address().port)
-        // using ES5 as Node.js <=4.0.0 does not have Computed Property Names
-        var reqHeaders = {}
-        reqHeaders[http2.constants.HTTP2_HEADER_PATH] = '/'
-        var request = client.request(reqHeaders)
+        var request = client.request()
         request.on('response', function (headers) {
           assert.strictEqual(headers[http2.constants.HTTP2_HEADER_STATUS], 200)
           assert.strictEqual(headers[http2.constants.HTTP2_HEADER_CONTENT_TYPE], 'text/plain')
--- a/test/session.js
+++ b/test/session.js
@@ -16,6 +16,16 @@ var Cookie = require('../session/cookie')
 
 var min = 60 * 1000;
 
+var describeHttp2 = describe.skip
+try {
+  var http2 = require('http2')
+  describeHttp2 = describe
+} catch (err) {
+  if (err) {
+    console.log('http2 tests disabled.')
+  }
+}
+
 describe('session()', function(){
   it('should export constructors', function(){
     assert.strictEqual(typeof session.Session, 'function')
@@ -2294,6 +2304,37 @@ describe('session()', function(){
       })
     })
   })
+
+  describeHttp2('http2', function () {
+    it('should work with http2 server', function (done) {
+      var server = createHttp2Server(null, function (req, res) {
+        res.setHeader(http2.constants.HTTP2_HEADER_CONTENT_TYPE, 'text/plain')
+        res.end('Hello, world!')
+      })
+      server.on('listening', function () {
+        var client = createHttp2Client(server.address().port)
+        // using ES5 as Node.js <=4.0.0 does not have Computed Property Names
+        var reqHeaders = {}
+        reqHeaders[http2.constants.HTTP2_HEADER_PATH] = '/'
+        var request = client.request(reqHeaders)
+        request.on('response', function (headers) {
+          assert.strictEqual(headers[http2.constants.HTTP2_HEADER_STATUS], 200)
+          assert.strictEqual(headers[http2.constants.HTTP2_HEADER_CONTENT_TYPE], 'text/plain')
+        })
+        var chunks = []
+        request.on('data', function (chunk) {
+          chunks.push(chunk)
+        })
+        request.on('end', function () {
+          closeHttp2(client, server, function () {
+            assert.strictEqual(Buffer.concat(chunks).toString(), 'Hello, world!')
+            done()
+          })
+        })
+        request.end()
+      })
+    })
+  })
 })
 
 function cookie(res) {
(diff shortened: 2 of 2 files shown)


test/session.js
```diff
@@ -2294,6 +2304,37 @@ describe('session()', function(){
...
+    it('should work with http2 server', function (done) {
+      var server = createHttp2Server(null, function (req, res) {
+        res.setHeader(http2.constants.HTTP2_HEADER_CONTENT_TYPE, 'text/plain')
+        res.end('Hello, world!')
+      })
+      server.on('listening', function () {
+        var client = createHttp2Client(server.address().port)
+        // using ES5 as Node.js <=4.0.0 does not have Computed Property Names
+        var reqHeaders = {}
+        reqHeaders[http2.constants.HTTP2_HEADER_PATH] = '/'
```
dougwilson, line 2318: It doesn't seem this is necessary, as https://nodejs.org/dist/latest-v18.x/docs/api/http2.html#clienthttp2sessionrequestheaders-options says that `/` is the default path (just like `GET` is the default method, and you're not specifying the method here).

dougwilson (changes requested):
I'm not sure what the purpose of the added test code is in relation to this change, as (1) reverting the changes in `index.js` doesn't affect the outcome of the tests and (2) looking at the test code it is just a http/2 server returning "Hello, world!" and asserting that that is the case...

test/session.js
lamweili: Good catch. Will remove it for consistency and less verbosity,

dougwilson:
Sweet. Also, I am going to sleep now, but meant to update here that I think I was using a Node.js version without http/2 earlier when I tested; I needed to get back and double-check that.

lamweili:
Ok, I have split the initial commit (d021958) into two separate commits (92d05cc, df7163e) for clarity.
1. 92d05cc - added test case which fails if `res._implicitHeader()` is used (e.g. in http2)
I added tests that actually require this change, such that we don't accidentally regress the behavior.
The existing codes are untouched. Checking out to it, we will get a test failure on existing codes.

2. df7163e - replaced `res._implicitHeader()` to address test case
In the next commit, which replaced the `res._implicitHeader()`, the test will now pass.

3. 6770cbf - removed explicit http2 headers for less verbosity (using default values)

I ran the GH actions on my forked repository and here are the results for all the commits:
https://github.com/lamweili/session/actions/workflows/ci.yml

lamweili:
I'm not sure if this change will provide/endorse full http2 support. 

IMHO, changing from an undocumented API to a documented API is a move in the right direction.
As a by-product, it helps to remove a blocker in the http2 scope, as http2 does not have `res._implicitHeader()`.

dougwilson:
Yep, makes sense. I assume the line right above that should be changed as well, right? `res._header` is also an undocumented HTTP API, which seems to be the purpose of the PR based on the description. We really should just only use public APIs for sure 👍 . A lot of this is just kruft, haha

lamweili:
Yup, agreed. 

This PR only fixes **ONE** undocumented API to documented API, which is `res._implicitHeader()`. 🤗

dougwilson:
Gotcha. Why spread it across a bunch of PRs? We can just fix the undocumented API usage at once. It's better for making a release, as making a release can cause folks to break. It is sad but true. I usually err on this being a semver major so we should get them all fixed at once.

lamweili:
Ah, I just did a carpet search in the repository, there is no usage for the other undocumented http API `res._header`.
So I reckon we got lucky?

dougwilson:
I'm not sure what that means. It is used in the line right above the one you changed in the PR. https://github.com/expressjs/session/blob/1010fadc2f071ddf2add94235d72224cf65159c6/index.js#L275

lamweili:
Ah, I will address the `res._header`  as well then. Any others to take note of?
https://github.com/expressjs/session/blob/1010fadc2f071ddf2add94235d72224cf65159c6/index.js#L275

dougwilson:
I'm on mobile, so not sure. I just saw that one right above the line you changed when looking at the diff. I can help take a look over the code to find them tomorrow 👍

lamweili:
Sure, thanks! While I have your attention, can you also kindly review https://github.com/expressjs/compression/pull/170?
(I will delete this comment thereafter, just needed the GitHub to notify you.)

dougwilson:
Yes, I will. I only had so much time today for open source. I was just doing them from top to bottom in the GH notifications. I haven't marked that notification as done yet, but also cannot promise that when I get back to it tomorrow, I'll look at that, esp. if you have new stuff on this one. I just go from top to bottom in my GH notifications with whatever time I have is all I can promise.

lamweili:
Continuation of https://github.com/expressjs/session/pull/908#issuecomment-1203407814:

4. c10aca5 - added tests to check for `res._header` and `res.headersSent` equivalence
Strangely, while [`res.headersSent` is only available from Node.js v0.9.4](https://nodejs.org/api/http.html#responseheaderssent), all test cases passed.
I was expecting the GitHub actions for Node.js v0.8 to fail. Perhaps the documentation is incorrect?

5. 647bdd1 - replaced `res._header` with `res.headersSent`

I ran the GH actions on my forked repository and here are the results for all the commits:
https://github.com/lamweili/session/actions

lamweili:
Continuation of https://github.com/expressjs/session/pull/908#issuecomment-1203407814:

4. 9e240a5 - added tests to check for `res._header` and `res.headersSent` equivalence
Strangely, while [`res.headersSent` is only available from Node.js v0.9.4](https://nodejs.org/api/http.html#responseheaderssent), all test cases passed.
Perhaps the documentation is incorrect?

5. e7b53e3 - replaced `res._header` with `res.headersSent` if available (prevent regression)

I ran the GH actions on my forked repository and here are the results for all the commits:
https://github.com/lamweili/session/actions

lamweili:
Continuation of https://github.com/expressjs/session/pull/908#issuecomment-1203407814:

4. 35880d6 - added tests to check for `res._header` and `res.headersSent` equivalence
Strangely, while [`res.headersSent` is only available from Node.js v0.9.4](https://nodejs.org/api/http.html#responseheaderssent), all test cases (including the added one) passed.
Perhaps the documentation is incorrect?

5. 7b8de6b - replaced `res._header` with `res.headersSent` if available (prevented regression so no need for major release)

I ran the GH actions on my forked repository and here are the results for all the commits:
https://github.com/lamweili/session/actions

lamweili:
Continuation of https://github.com/expressjs/session/pull/908#issuecomment-1203407814:

4. 35880d6 - added tests to check for `res._header` and `res.headersSent` equivalence
Strangely, while [`res.headersSent` is only available from Node.js v0.9.4](https://nodejs.org/api/http.html#responseheaderssent), all test cases (including the added one) passed.
Perhaps the documentation is incorrect?

5. 36618af - replaced `res._header` with `res.headersSent` if available (prevented regression so no need for semver major)

I ran the GH actions on my forked repository and here are the results for all the commits:
https://github.com/lamweili/session/actions

lamweili:
@dougwilson, have you got the time to take a look at this PR?

lamweili:
@dougwilson, have you got the time to take a look at this PR?

lamweili:
@dougwilson, have you got the time to take a look at this PR?

Martin-Luther:
I had the same issue : `http2` + `express-session` and the latest version of express-session `1.17.3` did not help.
`express-session` is __not the only module impacted__: I saw people having the same issue with `express-compression`.

> __NB.:__ I am not using `ExpressJS`, but my framework allows me to use modules/plugins from `ExpressJS`

__Option 1)__
You don't have the time to wait for a fix, you can intercept your response and add a test inside your script right before calling `res.end(...)`

```javascript
if (!res._implicitHeader) {
    // Hack required until `express-<plugin>` get support for http2 (express-session)
    res._implicitHeader = function(){ return; }; // Don't need, don't care
}
res.end();
```

This worked for me.

__Option 2)__
For those brave enough to fixe this issue for people like me.
Open `express-session/index.js` line @275

You should have something like:
```javascript
if (!res._header) {
    res._implicitHeader()
}
```

Needs to be replaced by:
```javascript
if (!res._header && res._implicitHeader) {
    res._implicitHeader()
}
```
This will skip the call for people using http2 or when `res._implicitHeader` not defined.

Let me know if it worked for you guys.

Martin-Luther:
I had the same issue : `http2` + `express-session` and the latest version of express-session `1.17.3` did not help.
`express-session` is __not the only module impacted__: I saw people having the same issue with `express-compression`.

> __NB.:__ I am not using `ExpressJS`, but my framework allows me to use modules/plugins from `ExpressJS`

__Option 1)__
You don't have the time to wait for a fix, you can intercept your response and add a test inside your script right before calling `res.end(...)`

```javascript
if (!res._implicitHeader) {
    // Hack required until `express-<plugin>` get support for http2 (express-session)
    res._implicitHeader = function(){ return; }; // Don't need, don't care
}
res.end();
```

This worked for me.

__Option 2)__
For those brave enough to fixe this issue for people like me.
Open `express-session/index.js` line @275

You should have something like:
```javascript
if (!res._header) {
    res._implicitHeader()
}
```

Needs to be replaced by:
```javascript
if (!res._header && res._implicitHeader) {
    res._implicitHeader()
}
```
This will skip the call for people using http2 or when `res._implicitHeader` not defined.

Let me know if it worked for you guys.

lamweili:
@dougwilson Not sure if you have the time to take a look again?

index.js
```diff
@@ -272,8 +272,8 @@ function session(options) {
           return ret;
         }
 
-        if (!res._header) {
-          res._implicitHeader()
+        if (typeof res.headersSent === 'boolean' ? !res.headersSent : !res._header) {
```
bjohansebas, line 275: Maybe it's better to make it a function like in compression https://github.com/expressjs/compression/pull/129

test/session.js
```diff
@@ -1371,6 +1381,25 @@ describe('session()', function(){
     })
   });
 
+  describe('res._header patch', function () {
```
bjohansebas, line 1384: I don't think this test makes sense.

bjohansebas:
closed by #1022

Issue #567, opened 2018-03-16
Use Session without cookies


sueess:
Is there any way to use this source or parts of it to create session without cookies? Or does anyone know a better fitting project?

I have a REST-api which is called by various services. The user-id, which I use as a session id, is in the payload of this requests. The incoming format of the JSON payload, and also the position of the user id, is different depending on the client, which sends the request.

I wanna use this user/session id to create a session, keep it alive and delete it after a certain time (clean up). The data could be kept in-memory or in a database.

 Many thanks for an advice.

wesleytodd:
I have a PR open for something like this (https://github.com/expressjs/session/issues/161).  Unfortunately, as it stands now the answer to your question is no.  Even with that PR it wouldn't support your use case.

One other note:  Using predictable session id's is an insecure practice and should be avoided.  So try not to use user id's like you described, but generate uuid's that are only used for auth and session tracking.  If you have to build your own solution, you can use how this package generates session id's as a resource.

Other Other note:  Instead of passing the session id in the body, it is more common to pass it in a header.  If it is also used for authentication, as is common, you can use the `Authorization` header.  Google for "bearer token http auth" or something like that for resources.

sueess:
Many thanks for that answer and your notes. I hoped there is a trick to use it this library without cookies. If anyone knows an alternative, please let me know.

knoxcard:
Close Ticket?

dotconnor:
You could implement a middleware before your session one that will find your client id in the body and then assign it to ```req.cookie['_sid']``` then everything should work as if it was a cookie.

vitalets:
This would be really useful for chatbots.

HarshithaKP:
@sueess Is the solution given by @dotconnor solved your problem or is this still occurring ?

sueess:
I developed my own session handling. I did not test the solution of @dotconnor, but it sounds feasible.

Issue #543, opened 2018-01-23
Cookieless Session


npshubh:
How can we implement cookie-less session in express js? Express-session provides a way to maintain session using cookies but what if we try to use express-session without cookies? Is there any way to achieve that?

dougwilson:
You could just call the `get` and `set` methods on your store object directly.

npshubh:
can you give me an example for that? How I'm supposed to do that?

dougwilson:
So when you create the store object using your chosen store module, that module has various methods to interact with it like `get`, `set`, and more. Depending on how you're going to handle management of sessions without cookies, you would interact with the store module. This module is basically a "cookie adapter" for all the various stores. You would need to implement your own adapter over the stores to provide the functionality you're looking for.

npshubh:
@dougwilson , when another user requests login using API, the first session stored in database is getting updated. So I'm losing session of previous user who was logged in using API. I'm unable to insert new session for each login request.

I can write custom code to achieve that but it worths to ask you if this library consists some logic to operate on such kind of scenario?

dougwilson:
Hm, I've never seen that happen here. It would be a big bug for sure. Can you help me reproduce the issue? I can take a look. Please provide all the following:

1. Version of Node.js
2. Version of this module
3. Complete server code that reproduces the issue
4. Instructions on how to get the server running if it is more than copy and paste your provided code
5. Instructions on how to reproduce in tge browser. Usually (1) go to this address (2) click this link, etc.

Thanks!

npshubh:
Node :- v9.2.1
"express-session": "^1.15.6",

Code is on my localhost so I can show you but I can explain you the process 👎 

1. I've used following code to setup 

app.use(session({
	secret: "test",
    cookies: { maxAge: 600000,httpOnly: false },
	saveUninitialized: false, // don't create session until something stored
	resave: false, //don't save session if unmodified
    store: new MongoStore({
    	mongooseConnection: mongoose.connection,
    	autoRemove: 'interval',
    	autoRemoveInterval: 10
    })
}));

2. Now whenever a user "A" goes to route "locahost/login" I'm setting a token "req.session.token = 'secret_token'; " and I get data in my mongodb database. Till now everything is fine, we have successfully created a session for user "A".

3. Now when another user "B" goes to route "locahost/login" and tries to login, then session of user "A" gets updated. So if user "A" tries to login using his token, then I can't find that in db and user "A" is not able to login.
[...]

Issue #317, opened 2016-05-26
Cookie less version?


olalonde:
I have a cookie-less API (which uses JWT for authentication) for a single page app and need a session store for the `passport-oauth1` module. Unfortunately, that module requires a session store to work. I was wondering if I could instead base a session store on a `req.query.session_id` query parameter instead of storing the session id in a cookie. Is there any module that behaves like `express-session` but will retrieve the session id from the url rather than a cookie?

gabeio:
There is a pr which will help do exactly what you are asking but we have not pulled it yet #159.

olalonde:
@gabeio thanks

gireeshpunathil:
what is the latest art on this? I don't believe we ever landed an RFE that will address this.

jilvin:
Is there someway we can support JWT instead of cookie? I was working on a project where we have to support clients other than web browsers and was looking to use JWT to store session id. I was wondering if there is some way I can use express-session together with JWT for the purpose.

dougwilson:
Hi @jilvin , can you not decode your JWT and just call the `.get` method on the store you're using to retrieve the session and `.set` to set it again?

jilvin:
@dougwilson Thanks for replying really fast. Appreciate it.

Yeah we were thinking of proceeding that way. But since this is one of the core part of our system as a whole we were thinking it would be better if we could use something really battle tested like `express-session`. We do not want to roll out creepy security bugs especially in something as critical as authentication. That's why I was trying to integrate `express-session` together with jwt instead of cookies.

I hope you might be able to provide more insights regarding this. Thanks in advance.

dougwilson:
Gotcha. So the main part of authentication / security would be in your JWT handling; a cookie-less version of this module wouldn't help in that regard, as securing your JWT would end up falling to you a implementation-specific details. For example, JWTs can be constructed in any way, signed or not, encrypted or not, come in anywhere in a request, etc.

The main features this module provides on top of get/set on the session store in management mainly around cookies specifically -- that when a cookie does not exist, it makes a new one, creates a new session, etc. Typically with these "cookie-less" flows like JWT, a request without a JWT is not just going to want to get a session created for it, it would typically be out-right rejected.

[...]

Issue #161, opened 2015-05-24
Sessions In API's


wesleytodd:
**Note:** So I am sorry if this is a duplicate of something else, but most of the issues I found relating to this are not all that helpful.

Using the RedisStore, but this is not redis specific, I was planning on sharing my sessions between a bunch of API microservices and a web app.  Obviously I don't want to use cookies for these api's, I want to use the auth header.  But the same session will also be accessed from my web app, which will use cookies.  

To solve this I would like to generate a single token to act as both my cookie value and my auth header.  When my web app makes a request to an api it would just load the cookie value into the auth header.

At my work we have a session module that does this and it works great, but it is not open source.  I was hoping to get your opinion on having this module support that.  Do you like the idea?  Do you know of another module that already does this?  Is there a way to get this module to do that already that I am just stupidly missing?  I would be happy to get this working in a PR or make a proof of concept first, just let me know.

wesleytodd:
So, I looked into the code a bit, and it looks like we could pretty easily break the main body of this code out into two drivers, which would just be middleware generator functions:

- `Cookie`: Gets and sets session in a cookie, basically the code that is here
- `Header`: Gets from the header but provides a method for generating the token to return to the client on authentication

This is basically the architecture we use at work for our session module.

wesleytodd:
https://github.com/wesleytodd/session/tree/api-sessions

I was pretty easily able to get the test passing after moving most of the login out into a driver.  I'm going to try and get a header implementation working next.  

Even if this doesn't look like something you guys would want to merge, I might maintain a fork, as I have needed something like this a few times in the past.

dougwilson:
Yes, this is desired. I'm currently out for the holiday weekend where I am, but I wanted to give you that feedback and let you know you're not being ignored :)

wesleytodd:
Awesome!  I just got it working in my api, it is very rudimentary, and to really get something good the entire module will need some attention.  But on that branch I linked to above you can see my 2 commits.

The things that would need to change to make this implementation clean:

- The `req.session.cookie` value is a misnomer in that the header driver just uses it for the maxAge stuff.  Would be better to call it something like `res.session.config`.
-  The current header driver doesn't do any signing.  My thought here is that it is not common practice to sign bearer tokens, they are just sent as is.  Which makes the secret NOT required.  This would probably need discussion.
- The drivers would be better broken out into separate small modules.  But right now they are way too specific.  They could easily be made into more generic modules that other people could use as the basis for different session implementations.

[...]

PR #910, opened 2022-08-22
basic cookieless session functionality

SimonAM:
If flag in option is set, sets header "X-Access-Token" instead of "Set-Cookie". 

Same fix as in a [comment](https://github.com/expressjs/session/issues/543#issuecomment-850952139) in #543. 

Related to: 
#543 
#161 
#567 
#317

### DIFF
--- a/index.js
+++ b/index.js
@@ -80,6 +80,7 @@ var defer = typeof setImmediate === 'function'
  * @param {String|Array} [options.secret] Secret for signing session ID
  * @param {Object} [options.store=MemoryStore] Session store
  * @param {String} [options.unset]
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -114,6 +115,10 @@ function session(options) {
   // get the cookie signing secret
   var secret = opts.secret
 
+  // should the header be set as token instead of cookie
+  var shouldReplaceCookieWithToken = opts.shouldReplaceCookieWithToken
+
```
ajmeese7, line 120: It looks like you have an extra carriage return here, other than that it looks good to me 👍🏼

ajmeese7 (changes requested):

moshe5745:
This decoupling from cookies is a convenient thing.
Any progress with this?

Issue #989, opened 2024-06-20
express-session vulnerability since it is still using a very old cookie-signature version


andiclone:
## Vulnerability
express-session, even in the latest v1.18.0, is still using cookie-signature v1.0.7 which is over a year old and it has a 'sha1' vulnerability: https://owasp.org/Top10/A02_2021-Cryptographic_Failures/

## Problem
In my project I have this reported since over 5 months ago with the latest change in this package, but still no newer version has come out to fix this vulnerability

## Solution
Upgrade the dependency on cookie-signature to a newer version, ideally 1.2.1 where it changes the old sha1 standard to a much more secure and updated sha256

## Notes
This is my first time posting an issue here so if I'm missing something please let me know :)

lucianidev:
i created a fix, I will create a pr

PR #990, opened 2024-07-12
changed the hashing algorithm from sha1 to sha256 due to security problems

lucianidev:
hello,
in the issue #989 the user references a vulnerability in the hashing algorithm sha1. The hashing method resulted vulnerable to hash collision, making it insecure. I changed the hashing method from sha1 to sha256 making it more secure. Tell me about any problems. 
I hope you have a good day

Documentation:
[https://crypto.stackexchange.com/questions/48289/how-secure-is-sha1-what-are-the-chances-of-a-real-exploit](url)
[https://www.quora.com/How-secure-is-SHA1-What-are-the-chances-of-a-real-exploit](url)
[https://stackoverflow.com/questions/38038841/why-is-sha-1-considered-insecure](url)

### DIFF
--- a/index.js
+++ b/index.js
@@ -614,7 +614,7 @@ function hash(sess) {
 
   // hash
   return crypto
-    .createHash('sha1')
+    .createHash('sha256')
     .update(str, 'utf8')
     .digest('hex')
 }


jonchurch:
sha1 is being used as a fast and efficient hash to detect changes in the session object

It is not cryptographically relevant, nor used for signing purposes.

krko12345:
Hello team, is this going to be addressed at some point soon?
Session is being reported as vulnerability by monitoring tools because of its use of sha1.

bjohansebas:
One option would be to allow the user to choose which hashing algorithm to use

jonchurch:
@bjohansebas I wouldn't accept that change personally. Adding an additional config for internals of the library that consumers don't have to worry about is net negative IMO. 

I understand that some automated security tools flag SHA-1 as a general vulnerability, but in this specific case, it’s a false positive. Switching to SHA-256 would introduce unnecessary overhead without adding security benefits.

PR #997, opened 2024-10-08
dep: cookie@0.7.2

knolleary:
Updates cookie dependency to 0.7.2


HISTORY.md
```diff
@@ -1,3 +1,8 @@
+unreleased
+==========
+
+ * deps: cookie@0.7.2
```
UlisesGascon, line 4:
```suggestion

  * deps: cookie@0.7.2
    - Fix object assignment of `hasOwnProperty`
  * deps: cookie@0.7.1
    - Allow leading dot for domain
      - Although not permitted in the spec, some users expect this to work and user agents ignore the leading dot according to spec
    - Add fast path for `serialize` without options, use `obj.hasOwnProperty` when parsing
  * deps: cookie@0.7.0
    - perf: parse cookies ~10% faster
    - fix: narrow the validation of cookies to match RFC6265
    - fix: add `main` to `package.json` for rspack
```

I will include the full changelog

UlisesGascon (approved):
Thanks @knolleary :heart: !

PR #1000, opened 2024-10-31
fix(buffer): use node:buffer instead of safe-buffer

ymeskini:
Since express@5 will support a minimum of nodejs version 18, I think it would be safe to use `node:buffer`

Cc @wesleytodd @UlisesGascon

### DIFF
--- a/index.js
+++ b/index.js
@@ -13,7 +13,7 @@
  * @private
  */
 
-var Buffer = require('safe-buffer').Buffer
+var Buffer = require('node:buffer').Buffer
 var cookie = require('cookie');
 var crypto = require('crypto')
 var debug = require('debug')('express-session');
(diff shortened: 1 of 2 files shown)


bjohansebas:
This would require dropping support for older versions, because it still supports Node.js versions that have vulnerabilities with the buffer

index.js
```diff
@@ -13,7 +13,7 @@
  * @private
  */
 
-var Buffer = require('safe-buffer').Buffer
+var Buffer = require('node:buffer').Buffer
```
ymeskini, line 16:
```suggestion
```

bjohansebas (approved):

UlisesGascon:
Hey @ymeskini! Thanks for your PR!

The session library is still compatible with Node@0.8 so this change can't be merged until the next major (in case that we drop support for old Node.js versions).

So, I will move this PR to draft to avoid accidental merges :+1:

wesleytodd:
Additionally before this could be merged, we would need to also remove it from the `package.json`. But I agree, leaving it in draft until the next major (which may honestly be quite a while) is a good idea.

PR #1003, opened 2024-11-21
fix: Resolve test failure - Refresh server.crt with existing key extending expiry to Nov 21 03:28:10 2034 GMT

BaileyFirman:
Cert used for testing has expired, used openssl to refresh with extended expiry 10 years into the future
```
// Create new cert
openssl x509 -in server.crt -signkey server.key -days 3650 -out new_server.crt

// Verify
openssl x509 -in new_server.crt -text -noout

// Finally replace
mv new_server.crt server.crt
```
```
// Content of new cert
Certificate:
    Data:
        Version: 3 (0x2)
        Serial Number:
            89:4e:6b:34:a9:c2:19:80
        Signature Algorithm: sha256WithRSAEncryption
        Issuer: C=US, ST=Indiana, O=node-express-session, CN=express-session.local
        Validity
            Not Before: Nov 21 03:28:10 2024 GMT
            Not After : Nov 19 03:28:10 2034 GMT
        Subject: C=US, ST=Indiana, O=node-express-session, CN=express-session.local
        Subject Public Key Info:
            Public Key Algorithm: rsaEncryption
                Public-Key: (1024 bit)
                Modulus:
                    00:d8:1b:7f:7c:ce:a3:7a:62:ff:c5:03:1e:fb:7a:
                    82:cb:8d:ee:15:34:26:03:5d:47:22:58:1c:dc:79:
                    a6:79:bb:fa:ed:3f:f5:ec:85:34:16:0c:a1:5f:51:
                    b6:ff:92:69:e2:0d:ac:ec:09:ab:12:a0:db:14:0a:
                    b7:2a:40:f2:45:d6:31:77:17:d5:e3:b1:9c:5c:82:
                    00:da:6f:c3:00:6c:03:21:9e:12:5a:c7:69:77:c8:
                    f2:6c:5a:70:7f:f2:6a:d8:71:db:1d:40:d3:d7:27:
                    14:16:f4:c3:19:81:4c:73:ef:78:08:63:9f:eb:11:
                    e6:b2:53:da:30:81:c9:d5:71
                Exponent: 65537 (0x10001)
        X509v3 extensions:
            X509v3 Subject Key Identifier:
                EA:5F:27:09:C1:16:E8:4E:19:34:B8:1D:C2:77:7F:A8:48:BC:F5:7E
    Signature Algorithm: sha256WithRSAEncryption
    Signature Value:
        93:fc:4d:6b:64:6c:5e:99:96:79:17:ca:97:f5:2a:13:b1:cd:
[...]

PR #1015, opened 2025-01-22
feat: gencert script to regenerate the test ssl certs

wesleytodd:
Opening a PR to show what I meant in my comment in this thread @ctcpip https://github.com/expressjs/session/pull/1003#issuecomment-2607509639

I was just thinking we could codify the whole thing so we don't need to debate what to do next time this happens.


ctcpip:
thanks Wes!  I've gone one further and added a commit that will ensure this is never* something we need to bother with again.  Please take a look when you have a chance 🙏

* famous last words

also, taking a look at the CI failure github is complaining about an obsolete action

wesleytodd:
Even better! The upload artifact stuff can be fixed out of band with this I think. My guess is that we need to make some similar changes to the CI (like turn off fast fail) that we have done in the other repos.

ctcpip (approved):

ctcpip:
reports making it to coveralls:

Screenshot 2025-01-23 at 12 45 10 PM

PR #1016, opened 2025-01-26
Update README.md

gebbber:
Clarified advantages to requiring signed cookies in addition to the use of a high-entropy session ID

### DIFF
--- a/README.md
+++ b/README.md
@@ -313,8 +313,10 @@ would best be a random set of characters. A best practice may include:
   - Periodic updates of the secret, while ensuring the previous secret is in the
     array.
 
-Using a secret that cannot be guessed will reduce the ability to hijack a session to
-only guessing the session ID (as determined by the `genid` option).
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -313,8 +313,10 @@ would best be a random set of characters. A best practice may include:
   - Periodic updates of the secret, while ensuring the previous secret is in the
     array.
 
-Using a secret that cannot be guessed will reduce the ability to hijack a session to
-only guessing the session ID (as determined by the `genid` option).
+Although primary session security is derived from the use of a high-entropy session
+ID, the additional use of signed cookies avoids needing to validate the contents of
+the cookie, and eliminates the need to implement rate limiting to avoid brute force
+attempts.
```
bjohansebas: This definitely does not eliminate the need to implement rate limiting, and the contents of the cookie must always be validated, as its contents and the information it carries cannot be assumed to be trustworthy.

PR #1020, opened 2025-02-10
chore: drop nodejs < 18

bjohansebas:
Support for old Node.js is removed, and the express CI is copied :)


bjohansebas:
@UlisesGascon can you bring the latest changes from the master branch to v2?

jonchurch:
Ah yeah, I see it needs the commit from Chris to generate a fresh cert

bjohansebas:
Yep, only you can do it since I don't have the permissions :)

.github/workflows/ci.yml
```diff
@@ -1,236 +1,112 @@
...
+          npm config set loglevel error
+        shell: bash
+
+      - name: Install dependencies
+        run: npm install
+
+      - name: Output Node and NPM versions
+        run: |
+          echo "Node.js version: $(node -v)"
+          echo "NPM version: $(npm -v)"
```
Phillip9587, line 73: I think we can remove that step because `actions/setup-node` already prints environment info:




```suggestion

```

Phillip9587:
I have just left two minor comments.

Also, I wanted to check—do we need to run the CI on `windows-latest`? The previous workflow only ran on `ubuntu-latest`. I believe our policy is that lower-level packages only run CI on Linux. Could you confirm if this change is intentional?

Let me know what you think!

bjohansebas:
Well, this is not a lower-level package, but I don't think we want to run tests on Windows here.

Phillip9587:
@bjohansebas Can you rebase the branch with the latest commits from `v2`?

bjohansebas:
It should have the latest changes from that branch

.github/workflows/ci.yml
```diff
@@ -1,236 +1,105 @@
...
+        shell: bash
+        run: npm run test-ci
 
-    - name: Merge coverage reports
-      shell: bash
-      run: find ./coverage -name lcov.info -exec printf '-a %q\n' {} \; | xargs lcov -o ./coverage/lcov.info
+      - name: Upload code coverage
+        uses: actions/upload-artifact@v4
+        with:
+          name: coverage-node-${{ matrix.node-version }}-${{ matrix.os }}
```
Phillip9587, line 75:
```suggestion
          name: coverage-node-${{ matrix.node-version }}
```

Phillip9587 (changes requested):

UlisesGascon (approved):

bjohansebas:
I would like to move forward with some changes for this package.

bjohansebas:
cc @UlisesGascon Can you merge this please, or can you make me a committer, please?

bjohansebas:
This blocked #1025 and #1023.

PR #1021, opened 2025-02-10
refactor: replace custom defer function with setImmediate

bjohansebas:
Since we are removing support for old Node.js, this will no longer be necessary


Phillip9587 (approved):

raphendyr (commented):
Looks good. I verified that `defer` doesn't exists in the codebase anymore.

PR #1022, opened 2025-02-10
feat: removing http internals

bjohansebas:
Although this does not provide direct support for HTTP/2, it is always good to avoid using Node.js internal APIs.

closes #908


wesleytodd:
Hey! I guess I am added as a reviewer on these ones? I am trying to say focused on just the core `express` deps, so I am going to remove myself from this one as a reviewer.

PR #1025, opened 2025-02-13
feat: enforce secret requirement for session creation

bjohansebas:
Now it is mandatory to have a secret to create the session, a deprecated feature has been removed.

Context: 

That is from the early days of Connect (https://github.com/senchalabs/connect/pull/580). Initially, they used cookie-parser to secure cookies. Now, the package itself has the option to handle it on its own, so it is safe to remove it since this package is being decoupled from cookie-parser


index.js
```diff
@@ -207,7 +207,7 @@ function session(options) {
 
     // backwards compatibility for signed cookies
     // req.secret is passed from the cookie parser middleware
-    var secrets = secret || [req.secret];
+    var secrets = secret;
```
bjohansebas, line 210: I need to review this part because if I remove it, the tests still pass. It's likely that several tests need to be created if this is necessary.

bjohansebas: That is from the early days of Connect (https://github.com/senchalabs/connect/pull/580). Initially, they used *cookie-parser* to secure cookies. Now, the package itself has the option to handle it on its own, so it is safe to remove it since this package is being decoupled from *cookie-parser*.

test/session.js
```diff
@@ -35,20 +35,8 @@ describe('session()', function(){
...
-
-  it('should get secret from req.secret', function(done){
-    function setup (req) {
-      req.secret = 'keyboard cat'
-    }
-
-    request(createServer(setup, { secret: undefined }))
-    .get('/')
-    .expect(200, '', done)
+  it('should reject without secret', function(){
```
raphendyr, line 38: This is now a "configuration error", so should this test be under the `describe('secret option', ...` block?

index.js
```diff
@@ -135,16 +135,16 @@ function session(options) {
...
 
-  if (secret && !Array.isArray(secret)) {
-    secret = [secret];
+  if (secrets && !Array.isArray(secrets)) {
+    secrets = [secrets];
   }
 
-  if (!secret) {
-    deprecate('req.secret; provide secret option');
+  if (secrets === undefined) {
```
raphendyr, line 146: I'm not sure about changing the test here. What is it supposed to achieve? I might be missing some JS knowledge here though.

Should we catch some invalid cases, like `secret: null` here too or are those something we should not care?

A test for the null case and fix for it:

```diff
diff --git i/index.js w/index.js
index 64f462f..835202a 100644
--- i/index.js
+++ w/index.js
@@ -143,7 +143,7 @@ function session(options) {
     secrets = [secrets];
   }

-  if (secrets === undefined) {
+  if (!secrets) {
     throw new Error('secret option required for sessions');
   }

diff --git i/test/session.js w/test/session.js
index af96eb8..ed083bc 100644
--- i/test/session.js
+++ w/test/session.js
@@ -1186,6 +1186,10 @@ describe('session()', function(){
       assert.throws(createServer.bind(null, { secret: [] }), /secret option array/);
     })

+    it('should reject null value', function () {
+      assert.throws(createServer.bind(null, { secret: null }), /secret option required for sessions/);
+    })
+
     it('should sign and unsign with a string', function (done) {
       var server = createServer({ secret: 'awesome cat' }, function (req, res) {
         if (!req.session.user) {
```

raphendyr (commented):
I find the logic to bee correct. Some minor notions in comments.

test/session.js
bjohansebas: yep! good catch

test/session.js
```diff
@@ -1194,6 +1178,13 @@ describe('session()', function(){
   });
 
   describe('secret option', function () {
+    it('should reject without secret',function () {
+      for (const secret of [undefined, null, '', false]) {
```
UlisesGascon, line 1182: Currently the linter is not happy with this: `1182:12  error  Parsing error: The keyword 'const' is reserved` ([ref](https://github.com/expressjs/session/actions/runs/13886394068/job/38851676761?pr=1025)). We can update the linter rules in a separate PR or in this one. WDYT @bjohansebas ?

bjohansebas: Hmm, it would be better in another PR. Maybe we should wait for https://github.com/expressjs/discussions/issues/327 to be defined, but if it takes too long, then it would be better to just update it.

raphendyr: To just get `const` to work, the linter ecmascript target/env needs to be updated. Hence, these lines needs to be added: https://github.com/expressjs/express/blob/master/.eslintrc.yml#L2-L3

There is the goal to drop support for Node <18 and it would make sense to match linter rules based on that. However, it might be ok to update the linter before migrating code or other such changes.

Would it be ok to update this repos lint rules based on what is in the "main" repo and after that, make better rules based on the discussion?

raphendyr: Alternatively, change this to `var` for this PR and fix it later.

raphendyr (commented):
I like it. Just needs a decisions about `const` and lint.

test/session.js
raphendyr: Sure, I'll make a PR for that.

Locally, I have just been working with the following:

```diff
diff --git i/.eslintrc.yml w/.eslintrc.yml
index b6b9f62..dc194a0 100644
--- i/.eslintrc.yml
+++ w/.eslintrc.yml
@@ -6,6 +6,9 @@ plugins:
 overrides:
   - files: '**/*.md'
     processor: 'markdown/markdown'
+env:
+  es2022: true
+  node: true
 rules:
   eol-last: error
   eqeqeq: ["error", "always", { "null": "ignore" }]
```

raphendyr: I created a PR for that, see #1030

test/session.js
```diff
@@ -1194,6 +1178,13 @@ describe('session()', function(){
...
+        console.log(secret)
```
bjohansebas, line 1183:
```suggestion
```

raphendyr (commented):
I think it's perfect now!

UlisesGascon (approved):

Phillip9587 (approved):

Issue #1, opened 2014-02-16
cookies and keygrip


jonathanong:
the plan was to use https://github.com/jed/cookies for cookie parsing and getting. express could include it as well via proto or something.

https://github.com/senchalabs/connect/issues/382
https://github.com/visionmedia/express/issues/1139

not exactly sure what to do though because the old `cookieParser` is incompatible with the new `cookies` since they both use `req.cookies`.

we could just move that into a pending v2 or something. don't know. i feel like this whole lib is jank and should be rewritten anyways.

@visionmedia

UlisesGascon:
This issue has been open for a while, and it seems no volunteers are willing to work on it. Therefore, I will proceed to close it.

PR #1026, opened 2025-02-13
feat: remove read signedCookies and cookies

bjohansebas:
As part of removing deprecated features and dropping compatibility with *cookie-parse*

### DIFF
--- a/index.js
+++ b/index.js
@@ -559,37 +559,6 @@ function getcookie(req, name, secrets) {
     }
   }
 
-  // back-compat read from cookieParser() signedCookies data
-  if (!val && req.signedCookies) {
-    val = req.signedCookies[name];
-
-    if (val) {
-      deprecate('cookie should be available in req.headers.cookie');
-    }
-  }
-
-  // back-compat read from cookieParser() cookies data
-  if (!val && req.cookies) {
-    raw = req.cookies[name];
-
-    if (raw) {
-      if (raw.substr(0, 2) === 's:') {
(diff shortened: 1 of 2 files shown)


raphendyr (commented):
I find this good and clean remove. 

Going forward, would there be benefits to eventually remove `cookie-parser` dev dependency? At the time of this PR, I value these tests though.

raphendyr:
Side note. As these other cookie libraries are removed, then this feature request is not going to be implemented either, I think: #1.

bjohansebas:
> Going forward, would there be benefits to eventually remove cookie-parser dev dependency? At the time of this PR, I value these tests though.

Yep, at some point, maybe for version 3. I’m keeping the library to make sure it's working well with that dependency. Maybe at some point, when we have the examples repository, we can move those tests there.

> As these other cookie libraries are removed, then this feature request is not going to be implemented either, I think: https://github.com/expressjs/session/issues/1. 

I would like to remove the handling of cookies from this library and leave it to cookies, that's why I reopened it.

raphendyr:
>> As these other cookie libraries are removed, then this feature request is not going to be implemented either, I think: https://github.com/expressjs/session/issues/1.

> I would like to remove the handling of cookies from this library and leave it to cookies, that's why I reopened it.

I think I misunderstood the issue last time I read it. Now, it makes total sense. Yeah, cookie prosessing should be handled by some other library.

UlisesGascon (approved):

Issue #988, opened 2024-06-19
Feature request: Allow Cookie options (at least domain) based on incoming requests


robaca:
We have a service serving multiple tenants with different domain names in the same process.
We'd like to set a domain part in the session cookie based on the incoming request.

We already tried creating multiple middleware instance and calling them in a wrapper middleware based on hostname, but that fails as the used cookie options are not that of the called middleware instance, but that of the last instantiated middleware.

Issue #920, opened 2022-11-15
【Feature】I wish I could define the name by req


code-farmer-i:
## now

```js
session({ name: 'a string' })
```

## expect

```js
session({ name(req) { return 'a string' } })
```

dougwilson:
Hello, and thank you for your suggestion. The goal of the configuration for this middleware is to be as static as possible. Can you elaborate on the specific use-case you are trying to accomplish that would make use of this?

wodCZ:
@dougwilson hi, I have a use-case for dynamic configuration.

We have a back-end staging environment running, currently serving as both a staging endpoint for staging instance of the webapp, which fits `sameSite: 'strict', secure: true` cookie setup just fine.
But, we'd like the staging instance to also accept requests from front-end developers localhost, so they can use it for development.

I'd like to configure that if the request origin is localhost, then use `sameSite: 'none', strict: true` for the request.

Is there a workaround for such use-case? I'm not sure about what are the consequences for the staging environment if I set `sameSite: 'none', secure: true` cookie settings statically for all requests. 

Thanks!

thom-nic:
I'm looking for something similar; I have a device with embedded webserver and by default is available over http and https.  If the user accesses over https, we could use `secure` however switching back to http will always fail until they clear their browser cookies.  

One solution would be if I could have a different cookie name when the request is https, and also dynamically set `secure: true` under the same condition.  So not only would `name` need to be dynamic but potentially other options as well.

I'm wondering if I could wrap the session instance - actually *two* session instances - in my own middleware function that chooses one or the other based on `req.secure`...

raphendyr:
Bit old issue, but few ideas for both.

For @wodCZ's case. One idea is to se `sameSite: 'none'` always for this environment as there shouldn't be security issues. Alternatively, make a router middleware, which selects which of two instances of express-session to use based on the request. Sadly, I think there are some complications to use multiple express-sessions. I those get fixed in future.

For @thom-nic's case. I would consider creating two express servers, one for each port (http and another for https). This way you can fully manage what middlewares are in each server. If you want to expose all the same routes via both servers, then it should be easy with a single router instance. Al though, this setup also would allow you to tailor which routes are bound to which server.

thom-nic:
I created a wrapper middleware that chooses the correct cookie name based on context and effectively has two separate session instances underneath.

PR #1027, opened 2025-02-26
feat: add support to dynamic cookie generation

lincond:
This patch introduces support for dynamic cookie generation in express-session. Instead of only accepting a static object for the options.cookie parameter, the patch allows developers to supply a function that returns a cookie options object. This enables dynamic generation of cookie settings on a per-request basis.

**Changes Made:**

**- JSDoc Update:**
    The type for options.cookie has been updated from Object to Object|Function to reflect the new functionality

**- Dynamic Cookie Options Resolution:**
    In the session middleware, when generating a new session, the code now checks if cookieOptions is a function. If it is, it calls the function with the current request (req) as an argument to retrieve the cookie options. Otherwise, it uses the static options as before.
    
```js
req.session.cookie = new Cookie(
  typeof cookieOptions === 'function' ? cookieOptions(req) : cookieOptions
);
```
**- Path Mismatch Check:**
    The cookie path validation now also resolves the cookie options dynamically to ensure that the generated cookie options are used in the check.
    
```js
var resolvedCookieOptions = typeof cookieOptions === 'function' ? cookieOptions(req) : cookieOptions;
if (originalPath.indexOf(resolvedCookieOptions.path || '/') !== 0) {
  // Handle pathname mismatch
  debug('pathname mismatch');
  next();
  return;
}
```

**Motivation and Use Case:**

This change was made to address scenarios where the cookie configuration (e.g., the cookie path) needs to be determined dynamically based on the incoming request. For example, in multi-tenant applications where different URL paths should result in different cookie configurations, this enhancement allows the session middleware to generate the correct cookie settings on the fly.

**Testing:**

To test this change, you can set the options.cookie to a function that returns a cookie options object based on the request. Verify that:
- When a function is provided, it correctly generates the cookie settings for each request.
- When a static object is provided, the behavior remains unchanged.

Feel free to provide any feedback or suggestions for further improvements.

### DIFF
--- a/index.js
+++ b/index.js
@@ -70,7 +70,7 @@ var defer = typeof setImmediate === 'function'
  * Setup session store with the given `options`.
  *
  * @param {Object} [options]
- * @param {Object} [options.cookie] Options for cookie
+ * @param {Object|Function} [options.cookie] Options for cookie
  * @param {Function} [options.genid]
  * @param {String} [options.name=connect.sid] Session ID cookie name
  * @param {Boolean} [options.proxy]
@@ -158,7 +158,7 @@ function session(options) {
   store.generate = function(req){
     req.sessionID = generateId(req);
     req.session = new Session(req);
-    req.session.cookie = new Cookie(cookieOptions);
+    req.session.cookie = new Cookie(typeof cookieOptions === 'function' ? cookieOptions(req) : cookieOptions);
 
     if (cookieOptions.secure === 'auto') {
       req.session.cookie.secure = issecure(req, trustProxy);
@@ -193,7 +193,8 @@ function session(options) {
 
     // pathname mismatch
     var originalPath = parseUrl.original(req).pathname || '/'
-    if (originalPath.indexOf(cookieOptions.path || '/') !== 0) {
+    var resolvedCookieOptions = typeof cookieOptions === 'function' ? cookieOptions(req) : cookieOptions
+    if (originalPath.indexOf(resolvedCookieOptions.path || '/') !== 0) {
       debug('pathname mismatch')
       next()
       return


bjohansebas:
I like the idea, although for it to be accepted, you need to add tests to verify the behaviors.

lincond:
> I like the idea, although for it to be accepted, you need to add tests to verify the behaviors.

I'll work on it!

lincond:
@bjohansebas Added tests to validate dynamic cookie generation via a function for the cookie option.

raphendyr:
As an observation, this is targeting v1/master and there is now some work on v2 with breaking changes. While this is not such a thing, it still might be good to know.

I started to prototype idea to put the whole cookie stuff behind class/api/something, so that people could swap it over or reimplement for custom needs. The prototype is here: https://github.com/raphendyr/express-session/commit/85c9bd9a3088c636f379e2a029cbff4eba4b12d1

When working on that, I keep this PR's need in mind too (where this merged or not).

lincond:
@bjohansebas added an example of how to use the callback as cookie options to `README.md` in addition to the tests. If you think you need anything else, let me know

bjohansebas (approved):

bjohansebas:
I guess it could go into version 1 and be backported to the v2 branch if this gets merged, What do you think, @UlisesGascon?

README.md
```diff
@@ -49,6 +49,26 @@ For a list of stores, see [compatible session stores](#compatible-session-stores
...
+    var match = req.url.match(/^\/([^/]+)/);
+    return {
+      path: match ? '/' + match[1] : '/',
+      httpOnly: true,
+      secure: req.secure || false,
+      maxAge: 60000
+    }
+  }
+}))
+```
```
raphendyr, line 70: One minor/nonimportant discussion point. Do you @lincond have some other example? Is this the use you have in production? This question is not here to contradict the PR. I think we might provide better service to the library users, if we can guide them bit more.

In this example, I guess `req.secure` might make sense, assuming you need request based secure handling (using same middleware for https and http port).

In my mind the request path based difference could be better handled by adding a separate instance to different paths. I think the `.cookie.path` should be easier to use with that setup in the first place. I mean something like this:

```js
const store = ...;
app.use('/app1', session({ ..., store, cookie: { path: '/app1' } })
app.use('/app2', session({ ..., store, cookie: { path: '/app2' } })
```

Sure, I would prefer the path could be set once per line, not twice.

I guess once problem is nested cookies, where we want one specifically for `/foobar` and another for `/`, i.e., everything else. Although, the code that selects cookie from request, will need different cookie name for each, I think.

There is also issue #920 relating to this, specifically setting the `name` based on request. In that, I was thinking maybe defining multiple express servers per port, and thus providing different `secure:` value, might be suitable.

Would you like to expand on your use case? Specially, if my thinking here is wrong or misguided. Thanks!

lincond: @raphendyr The example you gave works well when you know in advance how many “apps” you have. In our case, we have thousands of tenants and each of them has a specific cookie path that can only be known at runtime by checking the path and querying the database for data on that tenant. 

In addition, it would be necessary to upload a new version of the application every time a new tenant was added, so we used a callback-based solution to determine at runtime which tenant is making the request and assign the cookie to the tenant's path so that the user's session data doesn't leak to another tenant if the user accesses their site.

If you have any more questions, just let me know :)

lincond:
Closes #988

Goryounov:
Hello! Could you please let me know when the merge is planned? It's a very useful feature that, by the way, will replace this [pull](https://github.com/expressjs/session/pull/744).

lincond:
Hi all,

I've just rebased my branch with the latest upstream/master to incorporate recent commits.

@UlisesGascon, could you please review this PR when you have a moment?

@bjohansebas, would you mind assigning this to Ulises?

Thanks!

PR #1029, opened 2025-03-14
ci: add test for windows

bjohansebas:
Related https://github.com/nodejs/citgm/issues/1095#issuecomment-2725155135


test/support/gencert.sh
bjohansebas, line 1: I discovered that the error is in this file

UlisesGascon: Maybe we can make it work with some tweaks:

```sh
#!/usr/bin/env bash

set -e

# Detect OS
OS="$(uname -s)"

# Define file paths (cross-platform compatible)
KEY_PATH="./test/fixtures/server.key"
CERT_PATH="./test/fixtures/server.crt"

# Generate the certificate
if [[ "$OS" == "Linux" || "$OS" == "Darwin" ]]; then
    echo "Running on Unix-based system ($OS)"
    openssl req -x509 -nodes -newkey rsa:2048 \
    -keyout "$KEY_PATH" \
    -out "$CERT_PATH" \
    -days 3650 \
    -subj "/C=US/ST=Illinois/L=Chicago/O=node-express-session/CN=express-session.local"
elif [[ "$OS" == "MINGW"* || "$OS" == "MSYS"* || "$OS" == "CYGWIN"* ]]; then
    echo "Running on Windows (Git Bash)"
    openssl req -x509 -nodes -newkey rsa:2048 \
    -keyout "$KEY_PATH" \
    -out "$CERT_PATH" \
    -days 3650 \
    -subj "/C=US/ST=Illinois/L=Chicago/O=node-express-session/CN=express-session.local"
else
    echo "Unsupported OS: $OS"
    exit 1
fi

echo "Certificate generated successfully at: $CERT_PATH"
```

test/support/gencert.sh
```diff
@@ -1,5 +1,12 @@
 #! /bin/sh
 set -ex
 
-openssl req -x509 -nodes -newkey rsa:2048 -keyout ./test/fixtures/server.key -out ./test/fixtures/server.crt -days 3650 \
--subj "/C=US/ST=Illinois/L=Chicago/O=node-express-session/CN=express-session.local"
+OS=$(uname)
+
+if [ "$OS" = "Linux" ] || [ "$OS" = "Darwin" ]; then
```
raphendyr, line 6: Would it be better to reverse this if? I mean, should we check that if the OS is Windows, then use the alternative command.

The CI currently tests in windows and linux, but BSD and other similar systems are supported as far as I know.

raphendyr: Easy switch case could be:

```sh
case "$(uname)" in
  MINGW*) ... ;;
  *) ... ;;
esac
```

test/support/gencert.sh
bjohansebas: it's works

bjohansebas:
Apparently there's an issue with npm. I have no idea why, but it shouldn't be related to this PR. However, it's something we need to fix if it's affecting the CI.

test/support/gencert.sh
```diff
@@ -1,5 +1,17 @@
 #! /bin/sh
```
raphendyr, line 1: Should the shebang be `#!/bin/bash` or macos friendly `#/usr/bin/env bash`?. The file uses bash features (double angle brackets and pattern matching), so indicating that here would be good.

package.json
```diff
@@ -39,7 +39,7 @@
   },
   "scripts": {
     "lint": "eslint . && node ./scripts/lint-readme.js",
-    "test": "./test/support/gencert.sh && mocha --require test/support/env --check-leaks --bail --no-exit --reporter spec test/",
+    "test": "bash -c 'test/support/gencert.sh' && mocha --require test/support/env --check-leaks --bail --no-exit --reporter spec test/",
```
raphendyr, line 42: The command here executes bash, which will then execute the script. As far as I get it, it still should be running with the `/bin/sh` shell. In other words, the `-c` is extra here, assuming the point was to use bash as shell interpreter.

Alternatively, the original `./test/support/gencert.sh` is fine, as long as you fix the shebang in the script itself.

test/support/gencert.sh
```diff
@@ -1,5 +1,17 @@
...
-set -ex
+set -e
+
+# Detect OS
+OS="$(uname -s)"
+
+# Define file paths
+KEY_PATH="./test/fixtures/server.key"
```
raphendyr, line 8: Shellcheck output:

```
KEY_PATH="./test/fixtures/server.key"
^------^ SC2034 (warning): KEY_PATH appears unused. Verify use (or export if used externally).
```

I guess these were before or were supposed to be used as part the command below.

test/support/gencert.sh
```diff
@@ -1,5 +1,17 @@
...
+CERT_PATH="./test/fixtures/server.crt"
```
raphendyr, line 9: Shellcheck output:

```
CERT_PATH="./test/fixtures/server.crt"
^-------^ SC2034 (warning): CERT_PATH appears unused. Verify use (or export if used externally).
```

test/support/gencert.sh
```diff
@@ -1,5 +1,17 @@
...
+SUBJ="/C=US/ST=Illinois/L=Chicago/O=node-express-session/CN=express-session.local"
+
+if [[ "$OS" == MINGW* || "$OS" == MSYS* ]]; then
```
raphendyr, line 12: Shellcheck output:

```
if [[ "$OS" == MINGW* || "$OS" == MSYS* ]]; then
   ^-- SC3010 (warning): In POSIX sh, [[ ]] is undefined.
```

See the shebang comment above.

raphendyr (commented):
Some shell scripting notes, including notes from shellcheck.

test/support/gencert.sh
bjohansebas: fixed

test/support/gencert.sh
bjohansebas: fixed

test/support/gencert.sh
bjohansebas: fixed

package.json
bjohansebas: With just Bash, it's already enough, but we do need to use that command for it to work on Windows, since that's the important part of this PR.

package.json
```diff
@@ -26,7 +26,7 @@
     "eslint-plugin-markdown": "3.0.1",
     "express": "4.17.3",
     "mocha": "10.2.0",
-    "nyc": "15.1.0",
```
bjohansebas, line 29: To fix a small issue with Windows during installation

.github/workflows/ci.yml
```diff
@@ -1,236 +1,198 @@
...
+
+      - name: Remove npm module(s) ${{ matrix.npm-rm }}
+        run: npm rm --silent --save-dev ${{ matrix.npm-rm }}
+        if: matrix.npm-rm != ''
+
+      - name: Install Node version specific dev deps
+        if: ${{ matrix.npm-i != '' }}
+        run: npm install --save-dev ${{ matrix.npm-i }}
+
+      - name: Remove eslint
```
bjohansebas, line 142: For some strange reason, we need to remove eslint. Either way, the linter only runs once, and it's in the previous job.

bjohansebas:
The CI passes 🥳, wow, i really struggled with the old versions of Node.js.

.github/workflows/ci.yml
bjohansebas: Nah, that's how the CI has always been, nothing new

package.json
```diff
@@ -39,7 +39,7 @@
   },
   "scripts": {
     "lint": "eslint . && node ./scripts/lint-readme.js",
-    "test": "./test/support/gencert.sh && mocha --require test/support/env --check-leaks --bail --no-exit --reporter spec test/",
+    "test": "bash test/support/gencert.sh && mocha --require test/support/env --check-leaks --reporter spec test/",
```
bjohansebas, line 42: And while we're at it, let's remove '--bail' so we can see which tests fail simultaneously instead of one by one

test/support/gencert.sh
bjohansebas: fixed

raphendyr (commented):
Looks good to me (specially the shell part). I verified on my macos with `npm run test` and `./test/support/gencert.sh`.

bjohansebas:
I just tried it now that I have Windows, and this doesn't work.

PR #1030, opened 2025-03-18
lint: upgrade linting environment to ECMAScript 2022

raphendyr:
This PR updates the eslint JavaScript linting environment to the 2022 standard. This is based on the wanted Node 18 support and aligns with data in [node.green](https://node.green/) and what is used in the main express repo currently, see [expessjs/express/.eslintrc.yml](https://github.com/expressjs/express/blob/82fc12a40b3e6694e9a2c9b1376e7548d95779f6/.eslintrc.yml).

In addition to setting the ES version, we set `node` environment. It's not needed, I think, but makes the file match more with the main repo. The documentation for that says "Node.js global variables and Node.js scoping".

See comments below about added rule and the rule we have, which doesn't exists in the main repo.

### DIFF
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -6,10 +6,14 @@ plugins:
 overrides:
   - files: '**/*.md'
     processor: 'markdown/markdown'
+env:
+  es2022: true
+  node: true
 rules:
   eol-last: error
   eqeqeq: ["error", "always", { "null": "ignore" }]
   indent: ["error", 2, { "MemberExpression": "off", "SwitchCase": 1 }]
   no-mixed-spaces-and-tabs: error
   no-trailing-spaces: error
+  no-unused-vars: [error, { vars: all, args: none, ignoreRestSiblings: true }]
   one-var: ["error", { "initialized": "never" }]
--- a/README.md
+++ b/README.md
@@ -18,7 +18,10 @@ $ npm install express-session
 ## API
 
 ```js
-var session = require('express-session')
+const session = require('express-session')
+
+const app = express()
+app.use(session(/* options */))
 ```
 
 ### session(options)


.eslintrc.yml
```diff
@@ -6,10 +6,14 @@ plugins:
...
+  es2022: true
+  node: true
 rules:
   eol-last: error
   eqeqeq: ["error", "always", { "null": "ignore" }]
   indent: ["error", 2, { "MemberExpression": "off", "SwitchCase": 1 }]
   no-mixed-spaces-and-tabs: error
   no-trailing-spaces: error
+  no-unused-vars: [error, { vars: all, args: none, ignoreRestSiblings: true }]
   one-var: ["error", { "initialized": "never" }]
```
raphendyr, line 19: This rule exists in this repo, but not in expressjs/express. I think it's ok. It disallows the following:

```js
var foo = 1,
     bar = 2;
```

and forces the following:

```js
var foo = 1;
var bar = 2;
```

README.md
```diff
@@ -18,7 +18,10 @@ $ npm install express-session
 ## API
 
 ```js
-var session = require('express-session')
```
raphendyr, line 21: This line yielded the following error.

```
.../express-session/README.md
  21:5  error  'session' is assigned a value but never used  no-unused-vars

✖ 1 problem (1 error, 0 warnings)
```

Instead of skipping this, I though it would be ok to expand this introduction usage a bit. What do you think?

README.md
```diff
@@ -18,7 +18,10 @@ $ npm install express-session
...
+const session = require('express-session')
```
raphendyr, line 21: This could use the `import .. from ..` style as in expressjs/express, but I think that could be done later. However, I thought `const` is good here now.

bjohansebas (approved):

README.md
UlisesGascon: Sounds good to me! :)

.eslintrc.yml
UlisesGascon: +1 to this :+1:

.eslintrc.yml
```diff
@@ -6,10 +6,14 @@ plugins:
...
+env:
+  es2022: true
+  node: true
 rules:
   eol-last: error
   eqeqeq: ["error", "always", { "null": "ignore" }]
   indent: ["error", 2, { "MemberExpression": "off", "SwitchCase": 1 }]
   no-mixed-spaces-and-tabs: error
   no-trailing-spaces: error
+  no-unused-vars: [error, { vars: all, args: none, ignoreRestSiblings: true }]
```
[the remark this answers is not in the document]
UlisesGascon: Yep!

UlisesGascon (approved):
Thanks for investing time on this @raphendyr :heart:!

PR #1031, opened 2025-03-18
Request for comments / direction: Add types into repo and type checking without converting to typescript

raphendyr:
This is my test to add types to the repo. I wanted to take care of the following

1. Ensure types are close to the code with hope they stay updated and correct
2. Types would be defined in a single place. That would be `.d.ts` file or JSDoc. Final `.d.ts` files could be generated when building the package.
3. Ability to validate types. For the purpose to keep types correct, but also to provide static type checking to the implementation.
4. Based on discussion, it seems full Typescript transformation is not possibly wanted. Hence, I tried to use the second best technology JSDocs.

Based on this work. I find that:

1. Using JSDoc for this purpose is complicated. It's amazing how well it works, but there are issues that limit the ability in some cases, here for the implements/extends `CookieOptions`.
2. Google results, documentation and how Typescript works for JSDoc is very limited. I would predict that there are limited amount of people who would be able to work with this.
3. Codebase has some cohesion problems and hacks, which I think would have been mitigated if type checking would have been here from the start. See `Store.createSession`, which used to workaround how `Cookie` implementation works. There are some error propagation issues too, which I'm not totally sure. Might be my own doing or I just triggered those. I commented out those test at this point.

**If people see benefits for static type checking**, then I think it would make more sense to use Typescript in the first place.

**If people only want to include types in the repo**, then it might be suitable to define types only for the external API. This notably loses the type checking. In addition, I'm not sure how to design type validation here or how to verify it's up-to-date.

What do you people think?

Do you find good things from the code?

Do you find something that makes you worry?

Note that I focused on a single file, the `cookie.js`. Moving forward would require more work, but I wanted to talk about this direction before spending any more time.


types/cookie.d.ts
raphendyr, line 1: Note, this file is generated by the `tsc` command (`npm run check-types`). Most notably, that doesn't copy `cookie-options.d.ts` over, thus this folder doesn't for packaging as is.

The type generation from JSDoc works quite well.

The JSDoc comments in this file are limited, but at least vscode can find the description part from the `CookieOptions` when hovering over `Cookie` properties. I'm not sure if JSDoc tool can do the same.

session/cookie.js
```diff
@@ -7,146 +7,183 @@
...
-  },
-
-  /**
-   * Return a serialized cookie string.
-   *
-   * @return {String}
-   * @api public
-   */
-
-  serialize: function(name, val){
```
raphendyr, line 138: I don't know if this is used by external Store implementations. In this repo, this was used only by the test code. The version in `index.js` calls `cookie.serialize` directly on purpose or by accident.

session/cookie.js
```diff
@@ -7,146 +7,183 @@
...
-  },
+      : this.expires
+  }
 
   /**
    * Return cookie data object.
    *
-   * @return {Object}
-   * @api private
+   * @return {cookie.CookieSerializeOptions}
```
raphendyr, line 149: This interface was and is still used for multiple purposes. The code isn't very clear who it's serving.

Specifying this type, it makes it clear it has a single purpose. As a result, the property could be renamed.

For now, I used this for `toJSON`, but it might be better to separate these two.

session/cookie.js
```diff
@@ -7,146 +7,183 @@
...
+      this.maxAge = options.maxAge
+      this.originalMaxAge ??= options.maxAge // FIXME: rethink this
+
+      this.partitioned = options.partitioned
+      this.priority = options.priority
+      this.secure = options.secure
+      this.httpOnly = options.httpOnly ?? true
+      this.domain = options.domain
+      this.path = options.path || '/'
+      this.sameSite = options.sameSite
```
raphendyr, line 66: Using a for loop here is not good from the type checking point of view (the for loop doesn't understand what type keys are).

The `Object.assign(this, options)` does work here, but then we need to skip `options.data` property some how. For some reason, people have been passing it here and it fails as `data` is readonly for `Cookie`.

As a good result, this fixes the problem with `data` field, which was special case. In addition, the code is quite clean.

As a down side, now these properties are hardcoded in quite many places and changes to those require care.

session/cookie.js
```diff
@@ -7,146 +7,183 @@
...
+
+      this.signed = options.signed // FIXME: how this is used??
```
raphendyr, line 68: For some reason there are undocumented properties in the type file. I'm didn't yet search what these were, why and for what purpose.

types/cookie.d.ts
raphendyr, line 1: Note, this file is generated by the `tsc` command (`npm run check-types`). Most notably, that doesn't copy `cookie-options.d.ts` over, thus this folder doesn't for packaging as is.

The type generation from JSDoc works quite well.

The JSDoc comments in this file are limited, but at least vscode can find the description part from the `CookieOptions` when hovering over `Cookie` properties. I'm not sure if JSDoc tool can do the same.

session/cookie.js
```diff
@@ -8,145 +8,181 @@
...
-  },
+      : this.expires
+  }
 
   /**
    * Return cookie data object.
    *
-   * @return {Object}
-   * @api private
+   * @return {CookieSerializeOptions}
```
raphendyr, line 148: This interface was and is still used for multiple purposes. The code isn't very clear who it's serving.

Specifying this type, it makes it clear it has a single purpose. As a result, the property could be renamed.

For now, I used this for `toJSON`, but it might be better to separate these two.

session/cookie.js
raphendyr: First commit removed this method, but the second is more about just adding types and converting to a class, hence this method is back. The above comment still applies, and this method probably could be removed.

session/cookie.js
raphendyr: For the second iteration, I used the old code, just added type overwrites, so that the for loop works. I think the outdated code is kind of better, but I prefer the less rewriting style at this point.

types/cookie.d.ts
raphendyr: This is fixed by using something else than `.d.ts` as the file. I went with `.types.ts` suffix, but this could be anything I think.

session/cookie.js
raphendyr: Second iteration went back to the current API, which is a mix. As a result, it now uses `object` as type.

session/session.js
```diff
@@ -7,137 +7,168 @@
...
- */
+  /**
+   * Reset `.maxAge` to `.originalMaxAge`.
+   *
+   * @this {Session & Partial<SessionData>}
+   * @return {Session} for chaining
+   * @public
+   */
+  resetMaxAge() {
+    if (this.cookie) {
```
raphendyr, line 93: This is a "fix` that is result of a type checking. The `Session.cookie` can be undefined. If I remember right, there is issue about some edge case when this happens.

At least, this happens when there is a code bug in `Session.constructor()`.

bjohansebas:
Hey, thanks for opening this issue. This is a bit on hold for now since it requires some more discussion (see https://github.com/expressjs/discussions/issues/192). I know having the types integrated is great, but it's not a top priority at the moment. However, I can review it soon since it's something I'm interested in.

raphendyr:
> This is a bit on hold for now since it requires some more discussion (see [expressjs/discussions#192](https://github.com/expressjs/discussions/issues/192)).

Thanks for reminding about that again. It's important discussion after all.

> I know having the types integrated is great, but it's not a top priority at the moment.

I know. I tried to leave it after first iteration, but I really liked types when I was working on my idea to move the header handling code behind some interface and came back to types to support that rework.

And because it's not priority, I wanted to open it as a discussion place with some concrete examples.

> However, I can review it soon since it's something I'm interested in.

I don't need it needs "review" with the goal to merge it, but feel free to read the changes and write your opinions and feelings freely. At this point, I think it's more valuable to have more opinions and development ideas. I (or someone) can then test those out. Maybe we finally get something that can be considered for merging.

Thank any how!

PR #1034, opened 2025-04-14
perf: 2x faster hash stringify

cesco69:
With huge session object the hash method become slower because iterate on each key/value of the object, just for remove the cookie property on the first level!

This PR make hash method 2/3x faster!

[...]

PR #1035, opened 2025-04-14
feat: add new option for custom hash

cesco69:
Alternative to https://github.com/expressjs/session/pull/1034 and also close https://github.com/expressjs/session/pull/990

Allow user to pass a custom hash method to to detect changes in the session object


bjohansebas:
At the moment, we do not want to add this option.

PR #1064, opened 2025-06-03
feat: update default value of resave option to false

bjohansebas:
I think it's better to set it to false instead of making this field required.

@UlisesGascon @raphendyr WDYT?

### DIFF
--- a/README.md
+++ b/README.md
@@ -249,8 +249,7 @@ parallel requests to your server and changes made to the session in one
 request may get overwritten when the other request ends, even if it made no
 changes (this behavior also depends on what store you're using).
 
-The default value is `true`, but using the default has been deprecated,
-as the default will change in the future. Please research into this setting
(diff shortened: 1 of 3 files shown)


UlisesGascon (approved):

raphendyr:
I think this is likely good. Kind of safe default settings are better than mandatory values in my mind.

I don't remember what I thought about this before. Saving the session often is really problematic due to race condition as stated in the doc:

> Depending on your store this may be necessary, but it can also create race conditions where a client makes two parallel requests to your server and changes made to the session in one request may get overwritten when the other request ends

It could be possible that this setting should be removed, and if the store needs this, it should implement a "fake" `.touch()` method, which would do the resave. However, that would be a change in the current interaction and thus would take time. Hence, I believe we need to keep this setting for now.

p.s. Sorry, I have had other priorities, thus I haven't had the time I wanted for this project.

PR #1075, opened 2025-08-10
chore: use node:buffer instead of safe-buffer

Snafuh:
use node:buffer instead of safe-buffer

Same change was done and shipped in express already a while ago 
https://github.com/expressjs/express/commit/c70197ad33057e88565bd7bde6454f9225ead6cc

### DIFF
--- a/index.js
+++ b/index.js
@@ -13,7 +13,7 @@
  * @private
  */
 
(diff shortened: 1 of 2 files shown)


Snafuh:
This was already done in the v2 branch. My bad!

PR #1114, opened 2026-03-28
fix: use RFC 6265 path matching for cookie path option

UlisesGascon:
## Main Changes
- Fix session middleware activating on unintended paths (e.g., `/admin` matching `/administrator`)
- Implement RFC 6265 section 5.1.4 path-match algorithm for cookie path comparison
- Replace simple `indexOf` prefix check with segment-boundary-aware matching
- Document RFC 6265 5.1.4 compliance in `cookie.path` section of README, noting the change since 1.19.1 and the prior behavior

## Notes
The previous implementation used `originalPath.indexOf(cookiePath) === 0` which matched any path starting with the cookie path string, regardless of segment boundaries. Per [RFC 6265 section 5.1.4](https://datatracker.ietf.org/doc/html/rfc6265#section-5.1.4), a cookie-path should only match if the request-path is identical, or the cookie-path is a prefix followed by a `/` character.

This is a semver-patch change. It corrects a bug in path matching behavior without changing the public API. Applications using correctly segmented paths (e.g., `/admin` matching `/admin/users`) are unaffected. Only the unintended matching of non-segment-boundary paths (e.g., `/admin` matching `/administrator`) is corrected.

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,6 +126,12 @@ More information about can be found in [the proposal](https://github.com/privacy
 Specifies the value for the `Path` `Set-Cookie`. By default, this is set to `'/'`, which
 is the root path of the domain.
 
+Since 1.19.1, path matching follows [RFC 6265 section 5.1.4][rfc-6265-5.1.4]. This means
+the session middleware will only activate when the request path is an exact match or falls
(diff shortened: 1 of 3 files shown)


metadust:
Happy to help!

PR #1124, opened 2026-05-30
fix: detect secure socket when connection is missing

raashish1601:
## Changes

- Update secure-request detection to fall back to `req.socket.encrypted` when `req.connection` is missing.
- Preserve existing `req.connection` precedence while supporting environments where only `req.socket` is available.
- Add regression test under `session()` cookie security behavior: treat request as secure when `connection` is missing but `socket.encrypted` is true.

## Validation

- `npm run lint`
- `npx mocha --require test/support/env --check-leaks --no-exit --reporter spec test/`


bjohansebas (approved):
LGTM, thanks!

PR #1125, opened 2026-05-30
fix: refresh cookie options for rolling sessions

raashish1601:
## Changes

- Ensure cookie options are re-applied for each request when `rolling: true`, so updated config is reflected for existing sessions.
- Extracted cookie creation into a shared helper for consistency between new and existing sessions.
- Added regression test for an existing session receiving updated cookie attributes/expiry on rolling requests.

## Validation

- node node_modules/mocha/bin/mocha --require test/support/env --check-leaks --no-exit --reporter spec --grep "rolling option" test/session.js
- node node_modules/mocha/bin/mocha --require test/support/env --check-leaks --no-exit --reporter spec test/session.js
- node node_modules/eslint/bin/eslint.js index.js test/session.js

### DIFF
--- a/index.js
+++ b/index.js
@@ -158,18 +158,24 @@ function session(options) {
   store.generate = function(req){
     req.sessionID = generateId(req);
     req.session = new Session(req);
-    req.session.cookie = new Cookie(typeof cookieOptions === 'function' ? cookieOptions(req) : cookieOptions);
+    req.session.cookie = createCookie(req);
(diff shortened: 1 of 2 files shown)


aprudnikoff:
Any plan to release it?

bjohansebas (commented):
I think this should be saved for a new major version. This could break applications, so we need to think this through more carefully.

PR #1131, opened 2026-06-28
deps: cookie@1.1.1

wojtekmaj:
Bumps `cookie` to 1.1.1 for the v2 branch, where the package already requires Node.js `>=18`.

This also updates express-session to use `cookie@1`'s current `parseCookie` and `stringifySetCookie` APIs instead of the deprecated `parse` and `serialize` aliases, and adjusts the invalid `expires` test expectation to match the more specific error message. This chase was not strictly necessary but reduces friction should we ever update `cookie` to v2.

Tests:
- `npm run lint`
- `npm test`

### DIFF
--- a/index.js
+++ b/index.js
@@ -521,7 +521,7 @@ function getcookie(req, name, secrets) {
 
   // read from cookie header
   if (header) {
-    var cookies = cookie.parse(header);
+    var cookies = cookie.parseCookie(header);
 
     raw = cookies[name];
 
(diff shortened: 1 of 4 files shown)


wojtekmaj:
> Bumps `cookie` to 1.1.1 for the v2 branch, where the package already requires Node.js `>=18`.
> 
> This also updates express-session to use `cookie@1`'s current `parseCookie` and `stringifySetCookie` APIs instead of the deprecated `parse` and `serialize` aliases, and adjusts the invalid `expires` test expectation to match the more specific error message. This change was not strictly necessary, but reduces friction should we ever update `cookie` to v2.
> 
> Tests:
> - `npm run lint`
> - `npm test`

bjohansebas:
I'll see if we can bump it to Node 22 and convert it to ESM. Thanks!

PR #1132, opened 2026-07-01
chore: add homepage and bugs URLs to package.json

DakshSinghDhami:
Adds `homepage` and `bugs.url` fields to package.json so users can easily find the project website and report issues.


bjohansebas:
They opened a large number of PRs, and they're basically all based on incorrect information. It's just spam.

PR #1135, opened 2026-07-06
feat: change default value of unset option from 'keep' to 'destroy'

bjohansebas:
Another TODO that was left over


UlisesGascon (approved):
LGTM! Also an important change to highlight 👍

PR #1138, opened 2026-07-06
feat: expire session cookie on destruction

bjohansebas:
Basically, we should expire the cookie after it's destroyed
Just as a reminder, the cookie must be exactly the same; otherwise, it won't expire.
supersedes https://github.com/expressjs/session/pull/242
closes https://github.com/expressjs/session/issues/241


UlisesGascon (approved):
LGMT! It is an important breakign change for v2 so probably we need to highlight it on release notes 👍

PR #1145, opened 2026-08-02
fix: avoid duplicate save after regenerate

### LINKED ISSUE #935: Regenerated session is re-saved even if not modified since save
PR #849 attempted to fix resaving an already-saved new session at the end of the request, but there's one corner case it missed. If `session.regenerate()` has been called before `session.save()`, then the request is still saved again at the end of the request.

This can actually lead to a race condition: if another request modifies the session after `session.save()` but before the end of the first request, then those modifications get overwritten when the first request ends. (And yes, this can happen in real world...)

How to reproduce: This code prints "saving" twice for one request (also verified by adding logging inside express-session)

```
const MemoryStore = require('express-session/session/memory')

const _set = MemoryStore.prototype.set
MemoryStore.prototype.set = function set() {
  console.log('saving')
  _set.apply(this, arguments)
}

const express = require('express')
const session = require('express-session')
const app = express()

app.use(session({
  secret: 'cat',
  resave: false,
  saveUninitialized: false
}))

app.get('/', (req, res) => {
  req.session.regenerate(() => {
    req.session.value = 'foo'
    req.session.save(() => {
      res.send('hello')
    })
  })
})

app.listen(3000)
```

deepakganesh78:
Fixes #935

## Reproduction
On current `master`, a request that calls `req.session.regenerate()`, mutates the new session, and then calls `req.session.save()` writes the same regenerated session twice. A minimal supertest reproduction counted `MemoryStore#set` calls and showed `total saves after response: 2`.

## Root cause
`regenerate()` replaces `req.session` with a new `Session`, but the middleware only wrapped the original session's methods. As a result, a manual `save()` on the regenerated session did not update the request's saved-session tracking hash, so the response finalizer treated it as unsaved and wrote it again.

## Fix
Wrap `regenerate()` like `reload()` so the replacement session gets the middleware method wrappers. Track the explicitly saved session id in a separate `savedId` closure variable alongside `savedHash`, and use that pair only for `isSaved()`.

Cookie-setting and modification semantics are unchanged: `originalId` is not mutated by `save()`, so session id rotation still makes `isModified()` true and still emits `Set-Cookie` for the new id even when the regenerated session has no data changes.

## Compatibility / security notes
This is a conservative behavior fix: explicit session regeneration and saving still persists and sends the new session cookie, but avoids an extra store write that could overwrite concurrent updates. No dependencies or public API are changed.

## Validation
- Original regression test fails without the fix: `0 passing, 1 failing`, with `2 !== 1` store writes.
- Added no-data-change regression test fails under the first version of this fix: `0 passing, 1 failing`, missing the expected `Set-Cookie` header.
- Targeted regenerate/save tests pass with this rework: `2 passing`.
- Full suite via Git Bash on Windows: `175 passing (3s)`.
- `npm run lint`: passed.
- Direct `npm test` on Windows fails before running tests because npm/cmd cannot execute `./test/support/gencert.sh` (`'.' is not recognized...`); reran the same script/test command through Git Bash with `MSYS_NO_PATHCONV=1` successfully.

### DIFF
--- a/index.js
+++ b/index.js
@@ -223,6 +223,7 @@ function session(options) {
     var originalHash;
     var originalId;
     var savedHash;
+    var savedId;
     var touched = false
 
     // expose store
@@ -383,6 +384,8 @@ function session(options) {
       store.generate(req);
       originalId = req.sessionID;
       originalHash = hash(req.session);
+      savedHash = undefined
+      savedId = undefined
       wrapmethods(req.session);
     }
 
@@ -394,6 +397,10 @@ function session(options) {
 
       if (!resaveSession) {
         savedHash = originalHash
+        savedId = req.sessionID
+      } else {
+        savedHash = undefined
+        savedId = undefined
       }
 
       wrapmethods(req.session)
@@ -411,8 +418,19 @@ function session(options) {
 
     // wrap session methods
     function wrapmethods(sess) {
+      var _regenerate = sess.regenerate
       var _reload = sess.reload
       var _save = sess.save;
+
+      function regenerate(callback) {
+        debug('regenerating %s', this.id)
+        var sess = this
+        _regenerate.call(this, function () {
+          savedHash = undefined
+          savedId = undefined
+          rewrapmethods(sess, callback).apply(this, arguments)
+        })
+      }
 
       function reload(callback) {
         debug('reloading %s', this.id)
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -411,9 +418,20 @@ function session(options) {
...
     // wrap session methods
     function wrapmethods(sess) {
+      var _regenerate = sess.regenerate
       var _reload = sess.reload
       var _save = sess.save;
 
+      function regenerate(callback) {
+        debug('regenerating %s', this.id)
+        var sess = this
+        _regenerate.call(this, function () {
```
kilisamemarisaaa, line 428: The delegated call needs to be returned here. `Session.prototype.regenerate()` is documented and implemented to return `this` for chaining, but this wrapper replaces the method and currently returns `undefined`. I reproduced the change with the same middleware request against base `d209c1f` (`ret === currentSession`) and head `b1e56d4` (`ret === undefined`). Please `return _regenerate.call(...)` and add a regression assertion for the return-value contract.

kilisamemarisaaa (commented):
I found one API compatibility regression in the new regenerate wrapper.

